PR #10, opened 2021-11-16
403 Add Google Tag Manager support

mikehalmamoj:
On behalf of Nathan Russell.

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -67,4 +67,5 @@ export default {
     },
   },
   domain: get('INGRESS_URL', 'http://localhost:3000', requiredInProduction),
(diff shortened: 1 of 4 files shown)


server/config.ts
```diff
@@ -67,4 +67,5 @@ export default {
     },
   },
   domain: get('INGRESS_URL', 'http://localhost:3000', requiredInProduction),
+  gtm_containerId: get('GOOGLE_TAG_MANAGER_CONTAINER_ID', null, requiredInProduction),
```
petergphillips, line 70: typescript variable with an underscore in the name :nauseated_face:

joelstobart-moj (approved):
LGTM

server/config.ts
petergphillips: and I think will fail as no helm configuration

PR #17, opened 2021-11-23
Remove memory store based cookie storage in favour of redis store

nathanrussell-moj-digital:
* Removed memory store based cookie storage in favour of redis store already setup in `setupWebSession.ts`
* Changed alert channels
* Setup session variable to hold auth token in prep for next PR

### DIFF
--- a/integration_tests/integration/link/requestLink.spec.ts
+++ b/integration_tests/integration/link/requestLink.spec.ts
@@ -9,14 +9,6 @@ context('Request Link Page', () => {
     cy.task('stubRequestLink')
   })
 
-  it('should render request link page to unauthenticated user and reset cookie', () => {
-    cy.setCookie('create_barcode_token', '... a barcode ...')
-    cy.visit('/link/request-link')
-    Page.verifyOnPage(RequestLinkPage)
-
-    cy.getCookie('create_barcode_token').should('not.exist')
-  })
-
   it('should render request link page without a sign out link', () => {
     cy.visit('/link/request-link')
     const requestLinkPage = Page.verifyOnPage(RequestLinkPage)
--- a/server/middleware/link/setupRequestLink.ts
+++ b/server/middleware/link/setupRequestLink.ts
@@ -1,27 +1,13 @@
 import express, { Router } from 'express'
-import session from 'express-session'
 import flash from 'connect-flash'
 import RequestLinkController from '../../routes/link/RequestLinkController'
-import config from '../../config'
 import MagicLinkService from '../../services/link/MagicLinkService'
 
 export default function setUpRequestLink(magicLinkService: MagicLinkService): Router {
   const router = express.Router()
   const requestLinksController = new RequestLinkController(magicLinkService)
 
-  // TODO This uses the same max session age and secret as for a logged in user. Consider separating these into different env vars.
-  router.use(
-    session({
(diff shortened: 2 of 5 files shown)


integration_tests/integration/link/requestLink.spec.ts
```diff
@@ -9,14 +9,6 @@ context('Request Link Page', () => {
     cy.task('stubRequestLink')
   })
 
-  it('should render request link page to unauthenticated user and reset cookie', () => {
```
nathanrussell-moj-digital, line 12: Removed this test because we will no longer store the auth token in a cookie (future dev work - at the moment we dont store it anywhere)
The new approach will be to store the auth token in the session, but this test is harder to write based on persisting/clearing stuff in the session as I'm not sure cypress has access to the server side session does it?

server/middleware/link/setupRequestLink.ts
```diff
@@ -1,27 +1,13 @@
...
 import RequestLinkController from '../../routes/link/RequestLinkController'
-import config from '../../config'
 import MagicLinkService from '../../services/link/MagicLinkService'
 
 export default function setUpRequestLink(magicLinkService: MagicLinkService): Router {
   const router = express.Router()
   const requestLinksController = new RequestLinkController(magicLinkService)
 
-  // TODO This uses the same max session age and secret as for a logged in user. Consider separating these into different env vars.
-  router.use(
```
nathanrussell-moj-digital, line 13: Setting up the session here is unnecessary as it is already setup in `setupWebSession.ts`, complete with using the redis store rather than a memory store.

server/middleware/link/setupRequestLink.ts
```diff
@@ -1,27 +1,13 @@
...
-    session({
-      cookie: { maxAge: config.session.expiryMinutes * 60000 },
-      store: new session.MemoryStore(),
-      saveUninitialized: true,
-      resave: true,
-      secret: config.session.secret,
-    })
-  )
   router.use(flash())
-  router.use(express.json())
```
nathanrussell-moj-digital, line 23: Removed these lines as they are already setup in `setupRequestParsing.ts`. Using flash isnt setup though, so I've left that for the time being, but it doesnt feel like something we want/need to setup for each controller/path setup?

integration_tests/integration/link/requestLink.spec.ts
mikehalmamoj: No - you might have to save that integration test for the end and do it via the sign out link.

nathanrussell-moj-digital: 'No' as in 'no, dont remove the test'? or No as in 'no, cypress cant see the session so we'll have to test this differently via the sign out like at the end' ?

server/middleware/link/setupRequestLink.ts
mikehalmamoj: I had to add both in the PoC to get it working - but that may be because of misconfiguration in other areas.

integration_tests/integration/link/requestLink.spec.ts
mikehalmamoj: ```No as in 'no, cypress cant see the session so we'll have to test this differently via the sign out like at the end' ```

PR #18, opened 2021-11-23
Added route, controller and service to verify a magic link

nathanrussell-moj-digital:
This PR introduces the ability for a user to click through a magic link, for that link to be verified and for the user to be authenticated by the JWT token returned by the API

Introduced basic Find Recipient page and form as the resultant page once the user clicks through a magic link
Added accessibility testing on all integration tests


helm_deploy/send-legal-mail-to-prisons/values.yaml
```diff
@@ -41,6 +41,7 @@ generic-service:
       SYSTEM_CLIENT_ID: "SYSTEM_CLIENT_ID"
       SYSTEM_CLIENT_SECRET: "SYSTEM_CLIENT_SECRET"
       SESSION_SECRET: "SESSION_SECRET"
+      BARCODE_TOKEN_PUBLIC_KEY: "BARCODE_TOKEN_PUBLIC_KEY"
```
mikehalmamoj, line 44: We are now sharing the public key with the API and so the secret name (the right hand side) is `JWT_PUBLIC_KEY`.

server/config.ts
```diff
@@ -34,6 +34,11 @@ export interface ApiConfig {
 export default {
   https: production,
   staticResourceCacheDuration: 20,
+  barcodeTokenPublicKey: get(
+    'BARCODE_TOKEN_PUBLIC_KEY',
+    '-----BEGIN PUBLIC KEY-----\nMIIBojANBgkqhkiG9w0BAQEFAAOCAY8AMIIBigKCAYEAnKf2/0ffidVro/KXXXobH98tm3qRAOLmk/jBB/CUw/13ITvd/l0ECiT8Dq+aUAUrHK+rRDk8iCuZMD3JYQRW1NUEw9zXG408DNHZo1alDo9EoP4KIL6gV8h2AdyV0Fra3guCGuuMzF+aRbIEHNDx/NWEjOXXibTrIxXPxbEiI4yfcjLyUiusE9T7ZHlaTEDNA4zaDcY98LJs5B17l/h3Q8RYwLGlZe1HJkhi/IxeYzX6+OQZdozFWRSmJDMwJ4oAdXOnE5jLl3q6IzKYWJxkt8z8ITcXzSWLMKEqkYxBS4eO5dsWMRK8x9njtOrj8I0jppNJIDt0dwjfldXhUmLIVQABW7TdAhJsAq/RY2K1NjoQvsQpzThjrkHGRkKIAClK0N4XccwKNiTmI/wMQ5ec/6sNJYEZko4wIUAGb4u1kNJtqupwErau6jL8ffkkSBQNzoQyI5WrTaRnwOrqkgHO/W0GisMNc+RC1z/KFtO9ybRFRB21/4cHCeH7NugdEHI7AgMBAAE=\n-----END PUBLIC KEY-----',
+    requiredInProduction
+  ),
```
mikehalmamoj, line 41: The secret stored in Kubernetes doesn't include the `BEGIN PUBLIC KEY` or `END PUBLIC KEY` blocks because we are sharing the public key with the API that generates the JWT and it doesn't need them. So we'll need to add those blocks somehow.

server/middleware/barcode/barcodeAuthorisationMiddleware.ts
```diff
@@ -0,0 +1,32 @@
...
+import { verify } from 'jsonwebtoken'
+import config from '../../config'
+import logger from '../../../logger'
+
+export default function barcodeAuthorisationMiddleware(): RequestHandler {
+  return (req, res, next) => {
+    res.locals.barcodeUser = undefined
+
+    if (!req.session.createBarcodeAuthToken) {
+      logger.warn('Attempt to access protected barcode resource without an auth token')
```
mikehalmamoj, line 12: If the users bookmark the barcode resource then we could see this warning a lot - so I'm not sure it needs to be logged, BAU?

server/middleware/barcode/barcodeAuthorisationMiddleware.ts
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    const token = req.session.createBarcodeAuthToken
+    return verify(
+      token,
+      config.barcodeTokenPublicKey,
+      { algorithms: ['RS256'] },
+      (err: VerifyErrors, payload: JwtPayload) => {
+        if (err) {
+          logger.warn('Unable to verify auth token', token, err.message)
```
mikehalmamoj, line 23: If the error is that the token has expired then this is BAU - so not sure we need to warn. For any other errors - such as signed incorrectly - this is very interesting but might also be nice to see the payload to give us a clue as to where the bogus token came from?

server/routes/link/VerifyLinkController.ts
```diff
@@ -0,0 +1,26 @@
...
+import MagicLinkService from '../../services/link/MagicLinkService'
+import logger from '../../../logger'
+
+export default class VerifyLinkController {
+  constructor(private readonly magicLinkService: MagicLinkService) {}
+
+  async verifyLink(req: Request, res: Response): Promise<void> {
+    const secret = req.query.secret as string
+    if (!secret) {
+      logger.warn('User attempted to verify a link without a secret on the query string')
```
mikehalmamoj, line 11: I'm not sure what we're going to do with this log message - it might just become noise. I think it would be good at the end of this Epic to break out another ticket about metrics to capture around the login journey. It might be good to be able to see how many obviously fraudulent attempts to sign in vs genuinue sign ins for example.

server/views/partials/header.njk
```diff
@@ -20,7 +20,7 @@
         {% if user %}
           <li class="hmpps-header__navigation__item">
             <a data-qa="manageDetails" class="hmpps-header__link" href="/account-details" data-test="manage-account-link" target="_blank" rel="noopener noreferrer">
-              <span data-qa=header-user-name>{{ user.displayName | initialiseName }}</span>
+              <span data-qa="header-user-name">{{ user.displayName | initialiseName }}</span>
```
mikehalmamoj, line 23: I'm not sure we need initialiseName as all of our user names will be email addresses.

mikehalmamoj (changes requested):
We need to deal with the secret name for the public key being different or this won't deploy.

server/views/partials/header.njk
nathanrussell-moj-digital: This line is the existing block of code that was already there, which is displayed when `user` is present. As I understand it `user` is populated via `populateCurrentUser`, and is essentially when the user has auth'ed and come in via DPS (I might have got some of the words/tech wrong here, but you get my gist). 
I think this will be used by our app when the Mailroom Staff authenticate

The only change I made to this line was to add the quotes around the attribute value, and I only did that for consistency with everything else. Happy to change it back?

helm_deploy/send-legal-mail-to-prisons/values.yaml
nathanrussell-moj-digital: So I need to change this to:
```
BARCODE_TOKEN_PUBLIC_KEY: "JWT_PUBLIC_KEY"
```
Is that right?

server/middleware/barcode/barcodeAuthorisationMiddleware.ts
nathanrussell-moj-digital: Thats a fair shout - I'll remove that

server/routes/link/VerifyLinkController.ts
nathanrussell-moj-digital: OK, I'll remove it for now

server/views/partials/header.njk
mikehalmamoj: Oh hang on that's not true - some will be DPS users!

mikehalmamoj: Sorry missed your comment before my 2nd one - yes it's fine, my point is rubbish.

server/middleware/barcode/barcodeAuthorisationMiddleware.ts
nathanrussell-moj-digital: I've removed the log statement for now (as per other logging comments) . As you say, we can re-address logging and metrics in a subsequent ticket 👍

mikehalmamoj (approved):

PR #30, opened 2021-12-01
SLM-11 Fixes

mikehalmamoj:
* Turn off HTML5 validation - it's incomplete and gives the user a poor experience
* Show the generic email error message if the email length is too long
* Show errors next to the field as well as in the header list
* Tidy up client credentials for running locally


timharrison-moj (approved):
LGTM

PR #35, opened 2021-12-02
SLM-29: Added screen for Manual Barcode Entry

nathanrussell-moj-digital:
Added screen for Manual Barcode Entry (no form submission in this commit)

Screenshot 2021-12-02 at 15 00 31


assets/sass/components/_manual-barcode-form.scss
```diff
@@ -0,0 +1,34 @@
...
+    .govuk-form-group {
+      padding-right: 2rem;
+    }
+    .govuk-input {
+      text-align: center;
+    }
+  }
+
+  .example-barcode {
+    @media (min-width: 40.0625em) {
```
nathanrussell-moj-digital, line 18: Not sure if this is the best/correct value to use for a media query. I got the value from the rendered page in chrome dev tools from another element (that uses a GDS style), but I dont know if 40.0625em is what chrome is reporting it as after some computation/calculation; or if there is a sass variable I can use; or if its OK ?

There is no mention of using media queries on https://design-system.service.gov.uk/

server/views/partials/header.njk
```diff
@@ -1,7 +1,7 @@
 {% from "govuk/components/tag/macro.njk" import govukTag %}
 
-<header class="hmpps-header" role="banner">
-  <div class="hmpps-header__container">
+<header class="govuk-header" role="banner">
```
nathanrussell-moj-digital, line 3: I added these styles to the header so that the header bar gets the blue border at the bottom, as per the design team's designs.

mikehalmamoj (approved):

PR #38, opened 2021-12-07
SLM remove allow list

mikehalmamoj:
We're having issues with team members being blocked with 403s. I don't understand how the mesh of MOJ networks hangs together but we keep finding new and exciting ways to be blocked. The UI will be public anyway so just turning off the allow list completely.


andymarke (approved):
LGTM

PR #41, opened 2021-12-09
SLM-9: Added placeholder page for reporting problem manually entering barcode

nathanrussell-moj-digital:
Added placeholder page for reporting problem manually entering barcode
Also minor update to existing pages to fix minor accessibility problem


server/views/pages/barcode/find-recipient.njk
```diff
@@ -9,7 +9,7 @@
 
 {% block content %}
 
-  <main class="app-container govuk-body">
+  <section class="app-container govuk-body">
```
nathanrussell-moj-digital, line 12: This is the accessibility fix I applied to all pages - basically we had nested `<main>` elements which axe objected to (only a minor objection so it didnt fail our tests)
Anyway, the reason there were nested `<main>` elements is because `layout.njk` has the parent (and arguably the correct) `<main>` element, then each page had its own. I've simply changed these to `<section>` which feels semantically correct, and axe is happy again 👍

mikehalmamoj (approved):

PR #44, opened 2021-12-14
SLM-9: Plumbing in of barcode verification into controller, basic results page view

nathanrussell-moj-digital:


.eslintrc.json
```diff
@@ -54,7 +54,9 @@
             "semi": false,
             "arrowParens": "avoid"
           }
-        ]
+        ],
+        "no-shadow": "off",
```
nathanrussell-moj-digital, line 58: This allows typescript enums to be used without the IDE highlighting a linting error

server/views/pages/scan/scan-barcode-result.njk
```diff
@@ -0,0 +1,36 @@
...
+    {% if errors.length > 0 %}
+      {{ govukErrorSummary({
+        titleText: 'There is a problem',
+        errorList: errors,
+        attributes: { 'data-qa-errors': true }
+      }) }}
+    {% endif %}
+
+    <div class="govuk-grid-row">
+      <div class="govuk-grid-column-full">
```
nathanrussell-moj-digital, line 22: The view is deliberately basic. What I wanted to achieve in this PR was the service call and rendering a different heading on this page based on the API response (happy path vs. DUPLICATE)

The next PR will pretty the view up, adding the rest of the content etc

server/@types/forms/index.d.ts
```diff
@@ -9,5 +9,14 @@ declare module 'forms' {
...
     barcode?: string
+    status?: BarcodeStatus
+  }
+
+  export enum BarcodeStatus {
+    CREATED,
+    CHECKED,
+    DUPLICATE,
+    EXPIRED,
+    RANDOM_CHECK,
```
mikehalmamoj, line 20: It would be nice if this could be imported from the Swagger docs - I'll see what I can do.

server/routes/scan/ScanBarcodeController.ts
```diff
@@ -1,10 +1,80 @@
...
+import validate from './BarcodeEntryFormValidator'
+import ScanBarcodeService from '../../services/scan/ScanBarcodeService'
 
+/**
+ * Controller class responsible for scanning and verifying barcodes.
+ * In this context, 'scanning' means scanning with a handheld scanner device, and also manually entering the barcode
+ * number via a keyboard. The business process and service calls are the same, with the only real difference being
+ * the views that are rendered, so therefore this controller has methods for both scanning and manually entering
+ * barcodes.
+ */
```
mikehalmamoj, line 12: Any value to be had using inheritance here?

mikehalmamoj (approved):

server/@types/forms/index.d.ts
nathanrussell-moj-digital: Yes, that would be ideal. I copied it verbatim including its name and values from the API project, so if it could be part of the swagger spec and then imported as a type in the UI project that would be great.
I'll merge it as it is at the moment, and if we can get it in the swagger spec then we can refactor at that point 👍

server/routes/scan/ScanBarcodeController.ts
nathanrussell-moj-digital: Possibly, but lets see how subsequent changes to this class pan out over the next few PRs

PR #70, opened 2022-01-05
SLM-73: Basic prison selection dropdown

nathanrussell-moj-digital:
This PR adds a basic standard html `select` dropdown to the Find Recipient page for the prisons as returned from the Prison Register API.

There is quite a lot of wiring and setup in this PR (adding the new service, wiring into the controller, setting up mocks for the cypress tests etc) so this PR is deliberately basic in its view rendering (ie. not the accessible type ahead that we ultimately want), and it does not currently cache the prisons to prevent unnecessary and wasteful calls to the API. Both of these will come in subsequent PRs.


server/routes/barcode/FindRecipientView.ts
```diff
@@ -1,17 +1,30 @@
...
+    form: FindRecipientForm
+    errors: Array<Record<string, string>>
+    barcode: string
+    prisonRegister: Array<Record<string, string>>
+  } {
     return {
       form: this.findRecipientForm,
       errors: this.errors || [],
       barcode: this.barcode,
+      prisonRegister: this.prisonRegister.map(prison => {
```
nathanrussell-moj-digital, line 22: Maybe sorting the prison array by prison name would be sensible thing to do here, so that the view gets its drop down option sorted nicely (currently they are presented in the order that the service returns them from the API)

I'll probably do this in the next PR ...

mikehalmamoj (approved):

PR #72, opened 2022-01-06
SLM-73: Refactor anonymous GET method

nathanrussell-moj-digital:


server/services/prison/PrisonRegisterService.spec.ts
```diff
@@ -17,6 +17,16 @@ describe('Magic Link Service', () => {
   })
 
   describe('getActivePrisons', () => {
+    it('should make the request to the prison register without an authorization header', done => {
+      nock(config.apis.prisonRegister.url, {
+        badheaders: ['authorization'],
```
nathanrussell-moj-digital, line 22: Any headers in the request that are in the `badheaders` array will not match the request and will fail the test. This is perfect for what we want to do here 👍

server/data/restClient.ts
```diff
@@ -49,56 +49,35 @@ export default class RestClient {
...
+        if (err) logger.info(`Retry handler found API error with ${err.code} ${err.message}`)
+        return undefined // retry handler only for logging retries, not to influence retry logic
+      })
+      .query(query)
+      .set(headers)
+      .responseType(responseType)
+      .timeout(this.timeoutConfig())
+
+    if (this.token) {
+      request.auth(this.token, { type: 'bearer' })
```
nathanrussell-moj-digital, line 71: This logic (conditionally setting the `authorization` header) is tested in the new unit test I've introduced in this PR (ie. it tests that the header is not set), and also in the existing `hmppsAuthClient.test.ts` which has a test that explicitly asserts the `authorization` header is present with an expected value:
```
      fakeHmppsAuthApi
        .get('/api/user/me')
        .matchHeader('authorization', `Bearer ${token.access_token}`)
        .reply(200, response)
```

mikehalmamoj (approved):

PR #74, opened 2022-01-06
SLM-73: Added autocomplete for prison field

nathanrussell-moj-digital:
This PR changes the prison dropdown to an autocomplete / type-ahead if the user has JS enabled (progressive enhancement)
It also updates the docker-compose file to bring in the prison-register image when running locally, and it applys a sort order to the prison list

### DIFF
--- a/server/views/pages/barcode/find-recipient.njk
+++ b/server/views/pages/barcode/find-recipient.njk
@@ -32,7 +32,7 @@
           <p>Barcode number: {{ barcode }}</p>
           <div class="govuk-grid-row">
             <div class="govuk-grid-column-two-thirds">
-              <img alt="The generated barcode image" src="{{ barcodeImageUrl }}" id="barcodeImage" style="display: block">
+              <img alt="The generated barcode image" src="{{ barcodeImageUrl }}" id="barcodeImage">
             </div>
           </div>
         {% else %}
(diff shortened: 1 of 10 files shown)


server/views/pages/barcode/find-recipient.njk
```diff
@@ -32,7 +32,7 @@
           <p>Barcode number: {{ barcode }}</p>
           <div class="govuk-grid-row">
             <div class="govuk-grid-column-two-thirds">
-              <img alt="The generated barcode image" src="{{ barcodeImageUrl }}" id="barcodeImage" style="display: block">
+              <img alt="The generated barcode image" src="{{ barcodeImageUrl }}" id="barcodeImage">
```
nathanrussell-moj-digital, line 35: The use of inline styling causes chrome (and probably other browsers) to moan - it still works, but you get a console error.
Personally I think this element looks better if not display:block (ie. default of display:inline), but if we really want to make it block then we should use the GDS class `govuk-!-display-block`
eg:
```
<img alt="The generated barcode image" src="{{ barcodeImageUrl }}" id="barcodeImage" class="govuk-!-display-block">
```

mikehalmamoj: Yeah I spotted that after my PR was already in - but it's not staying on that page anyway so all good.

mikehalmamoj (approved):

PR #76, opened 2022-01-07
SLM-73: Use the PrisonRegisterStore to cache active prisons in PrisonRegisterService

nathanrussell-moj-digital:
This PR wires in the PrisonRegisterStore into the PrisonRegisterService, so that the active prisons are cached in the store (redis)

### DIFF
--- a/server/services/prison/PrisonRegisterService.ts
+++ b/server/services/prison/PrisonRegisterService.ts
@@ -2,7 +2,7 @@ import config from '../../config'
 import PrisonRegisterStore from '../../data/cache/PrisonRegisterStore'
 import RestClient from '../../data/restClient'
 import { PrisonDto } from '../../@types/prisonRegisterApiClientTypes'
-import { Prison } from './PrisonTypes'
+import { Prison } from '../../@types/prisonTypes'
 
 export default class PrisonRegisterService {
   constructor(private readonly prisonRegisterStore: PrisonRegisterStore) {}
--- a/server/services/prison/PrisonRegisterService.ts
+++ b/server/services/prison/PrisonRegisterService.ts
@@ -1,29 +1,37 @@
-import RestClient from '../../data/restClient'
 import config from '../../config'
+import PrisonRegisterStore from '../../data/cache/PrisonRegisterStore'
+import RestClient from '../../data/restClient'
 import { PrisonDto } from '../../@types/prisonRegisterApiClientTypes'
-
-export interface Prison {
-  id: string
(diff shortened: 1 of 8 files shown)


server/services/prison/PrisonRegisterService.ts
```diff
@@ -1,29 +1,37 @@
-import RestClient from '../../data/restClient'
 import config from '../../config'
+import PrisonRegisterStore from '../../data/cache/PrisonRegisterStore'
+import RestClient from '../../data/restClient'
 import { PrisonDto } from '../../@types/prisonRegisterApiClientTypes'
-
-export interface Prison {
```
nathanrussell-moj-digital, line 5: This has been moved into a source file of its own (see below), because otherwise it was introducing a circular dependency (ie. `PrisonRegisterService` depends on `PrisonRegisterStore`, but `PrisonRegisterStore` depended on `PrisonRegisterService` to get the `Prison` type)

server/services/prison/PrisonRegisterService.ts
```diff
@@ -1,29 +1,37 @@
...
-        (data as Array<PrisonDto>)
-          .filter(prison => prison.active === true)
-          .map(prisonDto => {
-            return {
-              id: prisonDto.prisonId,
-              name: prisonDto.prisonName,
-            }
-          })
-      )
+    try {
```
nathanrussell-moj-digital, line 15: Not sure how comfortable I feel about this and the asynchronous nature of it.
If it were java you'd make it `synchronized` in some way. I know that javascript is inherently single threaded and that node uses an event loop to prevent the single thread blocking.
I think its the concept of the event loop that is making me stop and think about this .... there is definitely scope for 2 requests to hit this method and for them both to get nothing back from redis and for them both to then request the data from the prison register and for them both to cache it.
In reality I know this will be rare (if ever) and if it does happen it wont do any harm. It just feels wrong ..... ?

mikehalmamoj: No harm no foul. Complicated synchronize logic doesn't seem worth the effort... and will make this even more complicated.

server/services/prison/PrisonTypes.ts
```diff
@@ -0,0 +1,4 @@
+export interface Prison {
+  id: string
+  name: string
+}
```
mikehalmamoj, line 4: Maybe this should live in the `@types` directory?

mikehalmamoj (approved):

PR #79, opened 2022-01-11
SLM-74 scan results page changes

nathanrussell-moj-digital:
This PR updates the Scan Results pages with the new content.
It does not replace the 'scan another barcode' form with the new 'Scan Another Barcode' button - that will be in the next PR. All this PR does is update the main content for each of the results pages - a few screenshots attached:

Screenshot 2022-01-11 at 10 21 39

Screenshot 2022-01-11 at 10 22 17

Screenshot 2022-01-11 at 10 23 47

Screenshot 2022-01-11 at 10 24 28

### DIFF
--- a/assets/sass/components/_scan-barcode-result.scss
+++ b/assets/sass/components/_scan-barcode-result.scss
@@ -1,0 +1,20 @@
+#scan-barcode-result-notification {
+
+  .govuk-notification-banner__header {
+    height: 25px;
+
+    h2.govuk-notification-banner__title {
+      @extend .govuk-visually-hidden
+    }
+  }
+
+  .success {
+    .govuk-notification-banner__content p {
+      // hook class to enable styling of success banner such as applying green tick
+    }
+  }
+
+  .notification {
+    // hook class to enable styling of notification banners
+  }
+}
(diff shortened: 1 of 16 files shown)


assets/sass/components/_scan-barcode-result.scss
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  .success {
+    .govuk-notification-banner__content p {
+      // hook class to enable styling of success banner such as applying green tick
+    }
+  }
+
+  .notification {
+    // hook class to enable styling of notification banners
```
nathanrussell-moj-digital, line 18: This allows us to style the notification banner if we get asked to (such as making it orange rather than blue without changing the markup or the accessible-ness of it :) ). For the time being I've left this here as a hook point for as and when we are asked to do it

server/views/partials/scan/scan-barcode-result-cannot-enter-barcode.njk
```diff
@@ -3,11 +3,16 @@
...
+    {% set html %}
+      <p>You need to investigate this item because the barcode can't be linked to an approved sender.</p>
+    {% endset %}
 
     {{ govukNotificationBanner({
-      text: 'Follow your usual procedure for checking Rule 39 mail.',
-      disableAutoFocus: true
+      html: html,
+      disableAutoFocus: true,
+      classes: 'notification barcode-doesnt-scan'
```
nathanrussell-moj-digital, line 15: Notification banners (blue by default) now have the class `notification` so we can target them with css ... and also a css class of the specific type if we get asked to style specific types of notification banner in a specific way (eg: barcode-doesnt-scan might need to red for example)

mikehalmamoj (approved):

PR #87, opened 2022-01-14
SLM-64: Review Recipients table

nathanrussell-moj-digital:
Screenshot 2022-01-14 at 10 33 32

### DIFF
--- a/server/routes/barcode/ReviewRecipientsView.ts
+++ b/server/routes/barcode/ReviewRecipientsView.ts
@@ -1,32 +1,15 @@
 import { Recipient } from '../../@types/prisonTypes'
 
-type TableCell = {
-  text?: string
-  html?: string
-}
-
 export default class ReviewRecipientsView {
   constructor(private readonly recipients: Array<Recipient>, private readonly errors?: Array<Record<string, string>>) {}
 
   get renderArgs(): {
-    recipients: Array<Array<TableCell>>
+    recipients: Array<Recipient>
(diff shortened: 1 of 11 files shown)


server/routes/barcode/ReviewRecipientsView.ts
```diff
@@ -1,15 +1,32 @@
...
+
+  private generateTableRows(): Array<Array<TableCell>> {
+    return this.recipients.map(recipient => {
+      return Array.of(
+        { text: recipient.prisonerName }, // Column 1 - Prisoner Name
+        { text: recipient.prisonNumber }, // Column 2 - Prison number or DOB - TODO SLM-81
+        { text: recipient.prisonAddress.premise }, // Column 3 - Prison name
+        { html: '<a href="" class="govuk-link">Edit details</a>' }, // Column 4 - Edit recipient details link - TODO SLM-63
+        { html: '<a href="" class="govuk-link">Remove</a>' } // Column 5 - Remove recipient link - TODO SLM-83
+      )
```
nathanrussell-moj-digital, line 29: I'm not entirely comfortable with us generating markup in code like this, or that the order of the cells in the rendered table is determined by this code (ie. markup and markup concerns are not in the .njk template), but I guess this is better than iterating through the `Recipient` array and diving into the various properties etc within the template. At least it can be tested here .... 🤷‍♂️

mikehalmamoj (approved):

PR #97, opened 2022-01-18
SLM-66: Choose envelope size

nathanrussell-moj-digital:
This PR implements the express request handler to allow the user to select their envelope size and to render the Print Coversheets page  with the correct content; including cypress tests


assets/sass/components/_select-envelope-size-form.scss
```diff
@@ -44,7 +44,7 @@
     }
   }
 
-  .govuk-label div.d4 {
```
nathanrussell-moj-digital, line 47: Correcting a type throughout where for some reason I had incorrectly referred to DL envelopes as D4 !

integration_tests/integration/barcode/selectEnvelopeSize.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+    Page.verifyOnPage(ChooseBarcodeOptionPage).continueToCoversheet()
+  })
+
+  it('should redisplay form with errors given no envelope size selected', () => {
+    const selectEnvelopeSizePage = Page.verifyOnPage(SelectEnvelopeSizePage)
+
+    selectEnvelopeSizePage.submitWithoutSelectingEnvelopeSize()
+
+    selectEnvelopeSizePage.hasErrorContaining('Select an option')
+    selectEnvelopeSizePage.signOut().click()
```
nathanrussell-moj-digital, line 30: We should probably look at this a bit more at some stage, but similar to what you had to do yesterday, I also had to sign out after each test because otherwise it remembers the session data 🤷‍♂️ 

Weirdly (and I'm not sure how yours works, but it does) .... the `Page.signOut()` method only selects the button, it doesnt actually click it !!!
We use the `signOut` method in a number of places and apparently dont click it!!, so I dont know what it does / how it works .... but I do know I needed it here 🤷

mikehalmamoj (approved):

PR #98, opened 2022-01-18
SLM-69 add images

mikehalmamoj:


server/@types/forms/index.d.ts
```diff
@@ -28,4 +28,10 @@ declare module 'forms' {
   export interface PdfForm {
     envelopeSize?: 'dl' | 'c4' | 'c5'
   }
+
+  export type RadioButtonOption = {
```
nathanrussell-moj-digital, line 32: I can see why you've moved this (so it can be reused by your new filter 👍 ), but I'm not sure its a 'form' type is it?
It feels like its a support class/type for some nunjucks filters. I wonder if it should be more local to the filters?

server/routes/barcode/PdfController.ts
```diff
@@ -19,7 +19,7 @@ export default class PdfController {
 
     req.session.pdfForm = { ...req.body }
     if (!validateEnvelopeSizeOption(req)) {
-      res.redirect('/barcode/pdf/select-envelope-size')
+      return res.redirect('/barcode/pdf/select-envelope-size')
```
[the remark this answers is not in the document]
nathanrussell-moj-digital: good spot! I dont know how that worked before, but I can see that it would be needed 👍

server/@types/forms/index.d.ts
mikehalmamoj: My thinking was: @types is where custom types live, I didn't fancy creating yet another type collection to hold a single type and forms seemed like the best place for it because the filter is used to populate a form. Not sure I agree on local to filters because that implies we should break up @types?

nathanrussell-moj-digital (approved):

PR #99, opened 2022-01-18
SLM-69 Prevent duplication of happy route to each page

mikehalmamoj:


integration_tests/integration/barcode/createBarcodeAddressImage.spec.ts
```diff
@@ -26,13 +14,11 @@ context('Create Barcode Image', () => {
       .barcodeAddressImageExists()
       .imageDownloadButtonExists('Gage-Hewitt-A1234BC.png')
       .imageCopyButtonExists()
-      .signOut()
```
mikehalmamoj, line 29: Did nothing! And not needed anyway.

integration_tests/pages/barcode/chooseBarcodeOption.ts
```diff
@@ -1,6 +1,8 @@
+/* eslint-disable import/no-cycle */
```
mikehalmamoj, line 1: This is a major downside - each page knows about both its previous and next pages resulting in a dependency cycle. On the plus side it only uses a static member from its previous page but this could encourage use of non-static members.

I guess the alternative is to have a God class that knows about all pages but that could get messy.

Any better ideas?

integration_tests/pages/barcode/createNewContact.ts
```diff
@@ -28,4 +30,8 @@ export default class CreateNewContactPage extends Page {
   prisonIdField = (): PageElement => cy.get('#prisonId')
 
   submitButton = (): PageElement => cy.get('#create-new-contact-form button')
+
+  static goToPage = (): CreateNewContactPage => FindRecipientByPrisonNumberPage.goToPage().happyPath()
+
+  happyPath = (): ReviewRecipientsPage => this.submitWithValidValues()
```
mikehalmamoj, line 36: Each page has a static `goToPage` function to get to the page, and a `happyPath` to get to the next page. This way all pages don't have to know about how to get to its previous page which prevents the duplication of the test setup logic.

integration_tests/pages/barcode/chooseBarcodeOption.ts
```diff
@@ -34,4 +36,10 @@ export default class ChooseBarcodeOptionPage extends Page {
   hasErrorContaining = (partialMessage: string): void => {
     cy.get('.govuk-error-summary__list').should('contain', partialMessage)
   }
+
+  static goToPage = (): ChooseBarcodeOptionPage => ReviewRecipientsPage.goToPage().happyPath()
+
+  happyPathImage = (): GenerateBarcodeImagePage => this.continueToImage()
+
+  happyPathCoversheet = (): SelectEnvelopeSizePage => this.continueToCoversheet()
```
mikehalmamoj, line 44: Some pages will need multiple `happyPath...` methods to model the various good ways through the page.

integration_tests/integration/barcode/selectEnvelopeSize.spec.ts
```diff
@@ -1,40 +1,22 @@
...
+    selectEnvelopeSizePage = SelectEnvelopeSizePage.goToPage()
   })
 
   it('should redisplay form with errors given no envelope size selected', () => {
-    const selectEnvelopeSizePage = Page.verifyOnPage(SelectEnvelopeSizePage)
-
     selectEnvelopeSizePage.submitWithoutSelectingEnvelopeSize()
 
     selectEnvelopeSizePage.hasErrorContaining('Select an option')
-    selectEnvelopeSizePage.signOut().click()
```
mikehalmamoj, line 30: These sign outs were not required... as suspected Cypress handles the sessions for us.

integration_tests/pages/barcode/selectEnvelopeSize.ts
```diff
@@ -41,4 +43,12 @@ export default class SelectEnvelopeSizePage extends Page {
   c4Radio = (): PageElement => cy.get('#select-envelope-size-form input[name=envelopeSize][value=c4]')
 
   submitButton = (): PageElement => cy.get('#select-envelope-size-form button')
+
+  static goToPage = (): SelectEnvelopeSizePage => ChooseBarcodeOptionPage.goToPage().happyPathCoversheet()
+
+  happyPathDl = (): PrintCoversheetsPage => this.submitHavingSelectedDlEnvelopeSize()
```
nathanrussell-moj-digital, line 49: Multiple happy paths could begin to look messy, as I think you mentioned earlier

nathanrussell-moj-digital:
This is an interesting idea that as you say has pros and cons.

I like the aim of reducing duplication of test setup - how to get to the start page for the given test class. And I like the fact that it means the test classes themselves now have no direct knowledge or interaction with cypress - in theory we could replace cypress with a.n.other page driver library and it should be an easier migration (ie. only the page classes know about how to interact with the page via the page driver library)

I do worry that things can become a bit too abstract and opaque - especially the happyPath style methods. I think the static `<PageClass>.gotoPage()` methods are OK, because its clear that we simply want to get to that page in order to start the test. But from reading the test class, reading a line that says `findRecipientPage.happyPath()` doesnt immediately tell me what that happy path is or what its doing. It makes the test very hard to read and reason about (especially if you are not the one who wrote it initially)

And as I think you mentioned, the happyPath method approach becomes a bit more difficult when you have several potential happy paths from that page (eg: selecting different envelope sizes and submitting)

Maybe we keep the static `gotoPage` type methods, but then use the methods in each page class to drive the behaviour in the test (ie. what we are doing at the moment). As long as those methods are well named (eg: `submitFormWithValidPrisonerNumber` etc), then reading the tests should be nice?

Just some thoughts

mikehalmamoj:
> This is an interesting idea that as you say has pros and cons.
> 
> I like the aim of reducing duplication of test setup - how to get to the start page for the given test class. And I like the fact that it means the test classes themselves now have no direct knowledge or interaction with cypress - in theory we could replace cypress with a.n.other page driver library and it should be an easier migration (ie. only the page classes know about how to interact with the page via the page driver library)
> 
> I do worry that things can become a bit too abstract and opaque - especially the happyPath style methods. I think the static `<PageClass>.gotoPage()` methods are OK, because its clear that we simply want to get to that page in order to start the test. But from reading the test class, reading a line that says `findRecipientPage.happyPath()` doesnt immediately tell me what that happy path is or what its doing. It makes the test very hard to read and reason about (especially if you are not the one who wrote it initially)
> 
> And as I think you mentioned, the happyPath method approach becomes a bit more difficult when you have several potential happy paths from that page (eg: selecting different envelope sizes and submitting)
> 
> Maybe we keep the static `gotoPage` type methods, but then use the methods in each page class to drive the behaviour in the test (ie. what we are doing at the moment). As long as those methods are well named (eg: `submitFormWithValidPrisonerNumber` etc), then reading the tests should be nice?
> 
> Just some thoughts

Agreed. As discussed I have removed the `happyPath` methods but kept the `goToPage` methods - let's see how we get on.

nathanrussell-moj-digital (approved):
LGTM

PR #100, opened 2022-01-19
SLM-66: Added PDF Renderer and wired in to route

nathanrussell-moj-digital:


server/data/gotenbergClient.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+const DEFAULT_PDF_OPTIONS: PdfOptions = {
+  marginTop: 0,
+  marginRight: 0,
+  marginBottom: 0,
+  marginLeft: 0,
+}
+
+export default class GotenbergClient {
```
nathanrussell-moj-digital, line 21: This class inspired heavily by the version in CVL 😁 (they are different though)

server/middleware/setupPdfRenderer.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/*
+ * This function accepts a Gotenberg client as its only argument.
+ * It returns a handler function to render a normal HTML view template to
+ * produce the raw HTML (including images, stylesheet etc). It then uses a
+ * callback function to pass rendered HTML view into the Gotenberg client
+ * to produce and return a PDF document.
+ */
+
+export default function setupPdfRenderer(client: GotenbergClient) {
```
nathanrussell-moj-digital, line 13: This function inspired heavily by the version in CVL 😁 (they are different though)

server/middleware/setupPdfRenderer.ts
```diff
@@ -0,0 +1,40 @@
...
+      options: { filename: string; pdfOptions?: PdfOptions } = { filename: 'document.pdf' }
+    ): void => {
+      res.render(view, pageData, async (error: Error, html: string) => {
+        if (error) {
+          throw error
+        }
+
+        res.header('Content-Type', 'application/pdf')
+        res.header('Content-Transfer-Encoding', 'binary')
+        res.header('Content-Disposition', `attachment; filename=${options.filename}`)
```
nathanrussell-moj-digital, line 27: `Content-Disposition: attachment; ....` causes the browser to download the file without displaying it
`Content-Disposition: inline; ....` causes the browser to render the file in the browser tab

Maybe we should make this configurable and part of the `options` object

server/routes/barcode/PdfController.test.ts
```diff
@@ -109,4 +115,23 @@ describe('PdfController', () => {
       expect(res.redirect).toHaveBeenCalledWith('/barcode/pdf/select-envelope-size')
     })
   })
+
+  describe('submitPrintCoverSheet', () => {
```
nathanrussell-moj-digital, line 119: Need to write some more happy path tests here once a few questions on SLM-66 have been answered (I have some time booked with Liora and Lauren to discuss)

server/routes/barcode/PdfController.ts
```diff
@@ -36,4 +40,34 @@ export default class PdfController {
...
+
+  async submitPrintCoverSheet(req: Request, res: Response): Promise<void> {
+    if (!req.session.recipients) {
+      return res.redirect('/barcode/find-recipient')
+    }
+    if (!req.session.pdfForm) {
+      return res.redirect('/barcode/pdf/select-envelope-size')
+    }
+
+    const barcodeImages = await Promise.all(
```
nathanrussell-moj-digital, line 52: This is the loop we will need in `GenerateBarcodeController` when we officially support multiple recipients
We basically need to loop through all recipients, calling `createBarcodeService.generateBarcode` on each one. The minor complexity is that `createBarcodeService.generateBarcode` returns a `Promise`, so we need to use `Promise.all` to wait for them all to be done.

And then there is the question of what happens if one of them fails .... should it fail them all, or just the one? ... One for another day ...

server/routes/barcode/PdfController.ts
```diff
@@ -36,4 +40,34 @@ export default class PdfController {
...
+          logger.error(`Could not generate a barcode for ${recipient.prisonNumber}, ${error}`)
+          return undefined
+        }
+      })
+    )
+
+    return res.renderPDF('pdf/barcode-cover-sheet', {
+      envelopeSize: req.session.pdfForm.envelopeSize,
+      barcodeImages,
+      printDebugInfo: true,
```
nathanrussell-moj-digital, line 70: Need to make `printDebugInfo` a config param / env var .... but it's useful being on for the time being ...

server/middleware/setupPdfRenderer.ts
mikehalmamoj: Could we make the filename of a similar format to the barcode image download - just think it will help if there are multiple prints in the print queue?

server/routes/barcode/PdfController.ts
mikehalmamoj: Hmmm one fails all fail doesn't sound great!

mikehalmamoj (approved):

server/middleware/setupPdfRenderer.ts
nathanrussell-moj-digital: Yes, that was my thinking, though not sure what to do if there are multiple barcodes in the 1 PDF (not that thats in scope for this ticket, but we know its coming)
Its one of the questions I have for Liora and Lauren ....

server/routes/barcode/PdfController.ts
nathanrussell-moj-digital: Indeed. At the moment I think my catch block will essentially skip the one that fails and will generate a PDF with all the others in it .... but we need to consider how we handle that to the user .... how do we tell them that their PDF contains n-1 barcodes? how do we tell them which one failed? how do we tell them that they all failed? (ie. they could all fail if there are multiple recipients, or it could just be the one recipient that fails)

In fact I'll add 'error handling' to my list of questions ....

PR #102, opened 2022-01-20
SLM-94: Fix prison name autocomplete when either blank or non-matching prison name is selected

nathanrussell-moj-digital:
This PR fixes SLM-94 which is the bug with autocomplete where blanking a previous valid valid, or changing a previous valid value to a non-matching value results in the previous value being submitted on the form.

As part of implementing the fix I have written a couple of cypress tests which assert the prison name field has the error highlight and message when in error, so I have also had to fix SLM-92 as part of this.

### DIFF
--- a/assets/js/page-enhancements.js
+++ b/assets/js/page-enhancements.js
@@ -30,6 +30,24 @@
         selectElement: element,
       })
     })
+    $('#create-new-contact-form').submit(event => {
+      // Before submitting the form set the prisonId select option to the value matching the autocomplete, or the null option if there is no match
+      // This is because the autocomplete currently only sets the select option when there is a valid match, which causes incorrect
+      // behaviour when the user performs a valid autocomplete, then changes it to blank or non-matching text. In this case the underlying
+      // select is not reset so we have to do it ourselves.
+      event.preventDefault()
+      const underlyingSelectField = $('#prisonId-select')
+      const autoCompleteField = $('#prisonId')
+      // In all cases start off by setting the underlying select option to the null/empty option
+      underlyingSelectField.children(`option[value='']`).prop('selected', true)
+      // Now find the option whose text matches the autocomplete field, and select it. If there is no match nothing will be selected
+      underlyingSelectField
+        .children('option')
+        .filter((idx, option) => option.text === autoCompleteField.val())
+        .prop('selected', true)
+      // Submit the form
+      event.target.submit()
+    })
   })
 })($, document)
 ;(enableCopyBarcodeButton = ($, document) => {
--- a/integration_tests/pages/barcode/createNewContact.ts
+++ b/integration_tests/pages/barcode/createNewContact.ts
@@ -14,15 +14,53 @@ export default class CreateNewContactPage extends Page {
   }
 
   submitWithValidValues = (): ReviewRecipientsPage => {
+    this.enterAValidPrisonerName()
+    this.typeAheadAValidPrison()
+    return this.submitForm(ReviewRecipientsPage)
+  }
+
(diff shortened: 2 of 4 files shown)


assets/js/page-enhancements.js
```diff
@@ -30,6 +30,24 @@
         selectElement: element,
       })
     })
+    $('#create-new-contact-form').submit(event => {
```
nathanrussell-moj-digital, line 33: I wouldn't normally comment to this degree, but this is such a nasty hack and the fact that we needed to do it in the first place left me feeling that we should document it for fear of someone inadvertently removing/changing this in the future (even though we do have the cypress tests introduced in this PR to protect us)

integration_tests/pages/barcode/createNewContact.ts
```diff
@@ -14,17 +14,55 @@ export default class CreateNewContactPage extends Page {
...
+    return this
+  }
+
+  typeAheadAValidPrison = (): CreateNewContactPage => {
+    this.clearPrisonField()
     this.prisonIdField().type('ashfield')
     this.pressEnterInPrisonIdField()
-    this.submitButton().click()
-    return Page.verifyOnPage(ReviewRecipientsPage)
+    return this
```
nathanrussell-moj-digital, line 32: Returning `this` rather than `Page.verifyOnPage(CreateNewContactPage)` is better in cases like this (where the page hasn't changed) because 

1. the page hasn't changed so we don't need to verify it
2. is (marginally) faster 
3. (most importantly 😄 ) the page class constructor (which `Page.verifyOn` uses) validates that the prisoner name field is auto focussed. This works when the page is loaded and rendered (ie. cases where `Page.verifyOn` is appropriate), but in this case we are simply interacting with the page and form fields. So the focussed field is not the prisoner name field

integration_tests/pages/barcode/createNewContact.ts
```diff
@@ -14,17 +14,55 @@ export default class CreateNewContactPage extends Page {
...
+  clearPrisonField = (): CreateNewContactPage => {
+    this.prisonIdField().clear()
+    return this
   }
 
   pressEnterInPrisonIdField = () => {
     this.prisonIdField().type('\n')
   }
 
+  submitForm<T>(T): T {
```
nathanrussell-moj-digital, line 51: I _think_ this is how you do generics in typescript. It means we can have one `submit` method, and you call it with the class name you expect to land on - eg: `submitForm(FindRecipientsPage)` or `submitForm(CreateContactPage)` etc.

As a separate improvement maybe we consider moving this `submitForm` method to the superclass (`Page`) and adding an abstract `submitButton` method to the `Page` class which all subclasses need to implement. That might be quite nice?

server/views/pages/barcode/create-new-contact.njk
```diff
@@ -53,7 +53,8 @@
               label: {
                 text: 'Select name of prison'
               },
-              items: prisonRegister
+              items: prisonRegister,
+              errorMessage: errors | findError('prisonId')
```
nathanrussell-moj-digital, line 57: This is the fix for SLM-92 (error highlight and message not appearing on prison name field) - I forgot to add it when I originally wrote the markup for the page 🙄 
We like easy fixes like this 👍 😁

integration_tests/pages/barcode/createNewContact.ts
mikehalmamoj: Agreed, maybe in another PR try it on the Page clas. Very nice!

mikehalmamoj (approved):

PR #105, opened 2022-01-20
SLM-66: Prevent new barcodes being created by re-printing the PDF

nathanrussell-moj-digital:


docker-compose.yml
```diff
@@ -60,6 +60,10 @@ services:
     container_name: gotenberg
     ports:
       - "3001:3000"
+    environment:
```
nathanrussell-moj-digital, line 63: I was getting timeout issues with gotenberg in my local docker compose env - given that we set these exact same env values in helm's `values.yml` I thought we'd better do it locally as well!

integration_tests/integration/barcode/selectEnvelopeSize.spec.ts
```diff
@@ -15,6 +15,8 @@ context('Select envelope size', () => {
   })
 
   it('should goto print coversheets given envelope size selected', () => {
+    cy.task('stubCreateBarcode')
```
nathanrussell-moj-digital, line 18: The barcode number is now generated slightly earlier in our process, so need this stub in this test

server/@types/prisonTypes.ts
```diff
@@ -17,4 +17,5 @@ export type Recipient = {
   prisonNumber: string
   prisonerName: string
   prisonAddress: PrisonAddress
+  barcodeValue?: string
```
nathanrussell-moj-digital, line 20: The basic idea is that once we have been through adding all the recipients, at the point of needed barcode numbers (ie. for the copy & paste route, and the PDF route), we have a service call (see below) that iterates through the recipients and calls the API for each one to give them a barcode number. Therefore the collection of recipients in the session will have the details of the prisoner and their prison, plus their barcode number

server/services/barcode/CreateBarcodeService.ts
```diff
@@ -42,28 +36,37 @@ export default class CreateBarcodeService {
...
-        barcode: barcodeValue,
-        barcodeImageDataUrl: this.generateAddressAndBarcodeImage(barcodeImage, recipient),
-      }
-    } catch (error) {
-      logger.error('Error generating barcode image', error)
-      throw error
-    }
-  }
-
-  private async createBarcode(token: string): Promise<string> {
```
nathanrussell-moj-digital, line 59: The `CreateBarcodeService` class now has 3 public methods on its API:
* generateBarcodeValue
* generateAddressAndBarcodeDataUrlImage
* addBarcodeValuesToRecipients

I'm not entirely convinced that addBarcodeValuesToRecipients lives in this class, but given what it does and how it does it I couldnt think of a neater/easier place to put it.

mikehalmamoj: This class already knows what a Recipient is so I think it's fine

mikehalmamoj (approved):

PR #107, opened 2022-01-21
SLM-66: Changes to enable downloading the coversheet PDF rather than open in a new tab

nathanrussell-moj-digital:


server/data/gotenbergClient.ts
```diff
@@ -1,25 +1,10 @@
 import superagent from 'superagent'
 import logger from '../../logger'
 import config from '../config'
-
-export type PdfOptions = {
```
nathanrussell-moj-digital, line 5: I have moved the `PdfOptions` type to the PDF Renderer rather than this gotenberg client class - on the basis that the gotenberg client isnt really anything to with PDFs per.se - its really just a REST client class that happens to know about the specifics of calling the gotenberg container.

server/middleware/setupPdfRenderer.ts
```diff
@@ -1,33 +1,49 @@
...
 import logger from '../../logger'
 
+export type PdfOptions = {
+  headerHtml?: string
+  footerHtml?: string
+  marginTop: number
+  marginBottom: number
+  marginLeft: number
+  marginRight: number
+  filename: string
```
nathanrussell-moj-digital, line 12: The filename and content disposition (whether its is a download (attachment) or to be displayed in a tab (inline)) are now able to be specified by the calling code 👍

server/views/pages/barcode/pdf/print-coversheets.njk
```diff
@@ -22,6 +22,9 @@
       <div class="govuk-grid-column-full">
         <h1 class="govuk-heading-xl">Print the coversheets</h1>
 
+<p>Your coversheet will download shortly. Check your downloads folder</p>
```
nathanrussell-moj-digital, line 25: The precise content for this page, and the button label on the "Select Envelope Size" page will be the subject of another ticket.

mikehalmamoj: 👍 Will this content need to change if JS is not enabled?

mikehalmamoj (approved):

server/views/pages/barcode/pdf/print-coversheets.njk
nathanrussell-moj-digital: As long as KH designs some appropriate content we'll be fine! 🙄 
Strictly speaking "your coversheet will download shortly" is a lie if you have JS disabled because it uses JS to click the link on page load. But TBH I think we could get away with it as its so minor.

PR #116, opened 2022-01-24
SLM-77: Map prison name into duplicate barcode response screen

nathanrussell-moj-digital:


integration_tests/mockApis/sendLegalMail.ts
```diff
@@ -192,9 +192,10 @@ const stubVerifyDuplicateBarcode = (): SuperAgentRequest =>
         status: 400,
         errorCode: {
           code: 'DUPLICATE',
-          userMessage: 'Someone scanned this barcode at 9:11 am on 8 December 2021 at LEI. It may be an illegal copy.',
+          userMessage:
+            'Someone scanned this barcode at 9:11 am on 8 December 2021 at HMP Altcourse. It may be an illegal copy.',
           scannedDate: '2021-12-08T09:11:23Z',
-          scannedLocation: 'LEI',
+          scannedLocation: 'ACI',
```
nathanrussell-moj-digital, line 198: Set our stub response to be ACI because our stub prison data only contains a few prisons, with LEI not being one of them

server/routes/scan/ScanBarcodeController.ts
```diff
@@ -99,14 +101,7 @@ export default class ScanBarcodeController {
         req.session.barcodeEntryForm.createdBy = checkBarcodeResponse.createdBy
       })
       .catch(errorResponse => {
-        const errorType = errorResponse.data?.errorCode?.code
```
nathanrussell-moj-digital, line 102: This catch block was getting messy (and was untested!) and was only going to get messier (and more untested!) with the logic to get the prison name for Duplicate barcode scans.
So I refactored it all out into a new class which I've lovingly called a 'Error Response Mapper', including writing unit tests for it 👍

server/services/prison/PrisonRegisterService.spec.ts
```diff
@@ -154,32 +152,17 @@ describe('Prison Register Service', () => {
     })
   })
 
-  function mockPrisonAddressData() {
```
nathanrussell-moj-digital, line 157: Even though this mock setup was called it was never used at runtime. Dont really understand why because it looks like it should have done, but I took the view that spending too long on it was a waste of time, and simply removed it 🤷‍♂️ 
The downside is that this test is tightly coupled with the contents of `prisonAddressData.json` ... but it was before I came along and removed it .... so .... 🤷‍♂️

server/services/prison/PrisonRegisterService.ts
```diff
@@ -38,6 +38,15 @@ export default class PrisonRegisterService {
       : Promise.reject(new Error(`PrisonAddress for prison ${prisonId} not found`))
   }
 
+  /**
+   * Returns the name of the prison (`premise` field from the PrisonAddress) for the specified prisonId
+   * or simply the prisonId if the prison cannot be found by it's ID
+   */
+  getPrisonNameOrId(prisonId: string): string {
```
nathanrussell-moj-digital, line 45: I added this to the service class, and deliberately decided to make it return the passed in prisonID if the corresponding prison was not found (rather than an async method that throws).
If it were an async method that throws then it means the calling code (in my case the new Error Response Mapper class) would need to mess about handling it .... and it just looked messy and horrible

mikehalmamoj (approved):

PR #124, opened 2022-01-25
SLM-78: Error handling for when barcode & image generation fails

nathanrussell-moj-digital:


integration_tests/mockApis/sendLegalMail.ts
```diff
@@ -306,6 +306,54 @@ const stubCreateBarcodeFailure = (): SuperAgentRequest =>
     },
   })
 
+const stubCreateBarcodeSuccessThenCreateBarcodeFailure = (): Promise<unknown> => {
+  return Promise.all([
+    stubFor({
+      scenarioName: 'createBarcodeThenFail',
```
nathanrussell-moj-digital, line 312: This is neat - you can setup a wiremock scenario (which is really just a collection of related requests), where each request has a required start state. The initial wiremock state is `Started`, and then you can trigger a new state when the request executes.
This means you can setup wiremock to respond to 2 successive calls to the same endpoint with different responses - perfect for saying "the call to generate a barcode value for the first recipient will work, but the call for the second recipient will fail" 👍

integration_tests/integration/barcode/createBarcodeAddressImage.spec.ts
```diff
@@ -21,4 +22,14 @@ context('Create Barcode Image', () => {
     const page = Page.verifyOnPage(ChooseBarcodeOptionPage)
     page.continueToImageErrors().hasErrorContaining('error')
   })
+
+  it('should show an error if creating the first barcode succeeds but subsequent barcodes fail', () => {
+    cy.task('stubCreateBarcodeSuccessThenCreateBarcodeFailure')
+    cy.go(-1) // back to Review Recipients in order to add a 2nd recipient
```
mikehalmamoj, line 28: Is the add recipient link also tested elsewhere?

nathanrussell-moj-digital: Yes, as part of `legalSenderJourneyE2e.spec.ts` 👍

.circleci/config.yml
```diff
@@ -114,7 +114,7 @@ jobs:
           command: curl -o wiremock.jar https://repo1.maven.org/maven2/com/github/tomakehurst/wiremock-standalone/2.27.1/wiremock-standalone-2.27.1.jar
       - run:
           name: Run wiremock
-          command: java -jar wiremock.jar --port 9091
+          command: java -jar wiremock.jar --port 9091 --jetty-acceptor-threads 1 --verbose
```
nathanrussell-moj-digital, line 117: This makes wiremock receive and process 1 request at a time (effectively single threaded), which prevents problems with the wiremock statemachine not working correctly (ie. the integration tests and the node app send requests to wiremock faster than it can process the first one including changing its state)

mikehalmamoj (approved):

PR #128, opened 2022-01-26
SLM-81 fixes

mikehalmamoj:
* Don't save prisoner name from previous form, it is no longer relevant
* Fix autofocus - errors first, or first field

### DIFF
--- a/assets/js/page-enhancements.js
+++ b/assets/js/page-enhancements.js
@@ -8,31 +8,6 @@
  * No code should be written here that renders a page non-functional if JS is not enabled or the script does not run/fails.
  */
 
-;(autoFocusBarcodeField = ($, document) => {
-  $(document).ready(() => {
-    $('#scan-barcode-form #barcode').focus()
-  })
-})($, document)
-;(autoFocusPrisonNumberField = ($, document) => {
-  $(document).ready(() => {
-    $('#find-recipient-by-prison-number-form #prisonNumber').focus()
-  })
-})($, document)
-;(autoFocusPrisonerNameField = ($, document) => {
-  $(document).ready(() => {
-    $('#find-recipient-by-prisoner-name-form #prisonerName').focus()
-  })
-})($, document)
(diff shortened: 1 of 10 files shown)


assets/js/page-enhancements.js
```diff
@@ -8,31 +8,6 @@
  * No code should be written here that renders a page non-functional if JS is not enabled or the script does not run/fails.
  */
 
-;(autoFocusBarcodeField = ($, document) => {
-  $(document).ready(() => {
-    $('#scan-barcode-form #barcode').focus()
-  })
```
mikehalmamoj, line 14: Moved all the autofocus stuff to the bottom - so the other enhancements that change page layout have a chance to do their thing first.

assets/js/page-enhancements.js
```diff
@@ -110,3 +85,58 @@
...
+  $(document).ready(() => {
+    $('#find-recipient-by-prison-number-form #prisonNumber').focus()
+  })
+})($, document)
+;(autoFocusPrisonerNameField = ($, document) => {
+  $(document).ready(() => {
+    $('#find-recipient-by-prisoner-name-form #prisonerName').focus()
+  })
+})($, document)
+;(autoFocusContactPrisonerNameField = ($, document) => {
```
nathanrussell-moj-digital, line 103: God, this autofocus stuff is getting out of hand. We cant keep doing this for every page - I bet the other apps dont do this?
I wonder if we should challenge the requirement to autofocus fields. Notwithstanding the possible accessibility issues and already proven UX issues, this just feels like the wrong thing to be doing

nathanrussell-moj-digital (approved):

assets/js/page-enhancements.js
mikehalmamoj: We could replace this with some JS that works for any screen - either focus on the first input field or the first input field in error - at least then we have consistency throughout the app. But I agree, it doesn't appear to have been thought out properly.

PR #134, opened 2022-01-31
SLM-109 save contact

mikehalmamoj:
* Attempt to create contacts
* Ignore if that fails (further design needed if errors should be heeded)
* Tidy up REST client for hmpps auth / slm tokens

### DIFF
--- a/integration_tests/integration/barcode/createBarcodeAddressImage.spec.ts
+++ b/integration_tests/integration/barcode/createBarcodeAddressImage.spec.ts
@@ -5,6 +5,7 @@ import ReviewRecipientsPage from '../../pages/barcode/reviewRecipients'
 
 context('Create Barcode Image', () => {
   beforeEach(() => {
+    cy.task('reset')
     ChooseBarcodeOptionPage.goToPage()
   })
 
--- a/integration_tests/integration/barcode/createNewContactByPrisonNumber.spec.ts
+++ b/integration_tests/integration/barcode/createNewContactByPrisonNumber.spec.ts
@@ -4,6 +4,10 @@ import ReviewRecipientsPage from '../../pages/barcode/reviewRecipients'
 import CreateNewContactByPrisonNumberPage from '../../pages/barcode/createNewContactByPrisonNumber'
 
 context('Create New Contact By Prison Number Page', () => {
+  beforeEach(() => {
+    cy.task('reset')
+  })
+
   it('should redirect to find-recipient given user navigates to Create New Contact without going via find-recipients first', () => {
     FindRecipientByPrisonNumberPage.goToPage()
     cy.visit('/barcode/find-recipient/create-new-contact')
--- a/integration_tests/mockApis/sendLegalMail.ts
+++ b/integration_tests/mockApis/sendLegalMail.ts
@@ -323,13 +323,13 @@ const stubCreateContact = (): SuperAgentRequest =>
       headers: {
         'Content-Type': 'application/json;charset=UTF-8',
       },
-      // body: `{
-      //   id: 1,
-      //   prisonerName: {{jsonPath request.body '$.prisonerName'}},
-      //   prisonId: {{jsonPath request.body '$.prisonId'}},
-      //   prisonNumber: {{jsonPath request.body '$.prisonNumber'}},
-      //   prisonerDob: {{jsonPath request.body '$.prisonerDob'}},
-      // }`,
+      body: `{
+        id: 1,
+        prisonerName: {{jsonPath request.body '$.prisonerName'}},
+        prisonId: {{jsonPath request.body '$.prisonId'}},
+        prisonNumber: {{jsonPath request.body '$.prisonNumber'}},
+        prisonerDob: {{jsonPath request.body '$.prisonerDob'}},
+      }`,
     },
   })
 
--- a/integration_tests/mockApis/sendLegalMail.ts
+++ b/integration_tests/mockApis/sendLegalMail.ts
@@ -306,6 +306,33 @@ const stubCreateBarcodeFailure = (): SuperAgentRequest =>
     },
   })
 
+const stubCreateContact = (): SuperAgentRequest =>
+  stubFor({
+    request: {
+      method: 'POST',
+      urlPattern: '/send-legal-mail/contact',
+      headers: {
+        'Create-Barcode-Token': {
+          equalTo:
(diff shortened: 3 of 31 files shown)


integration_tests/integration/barcode/createBarcodeAddressImage.spec.ts
```diff
@@ -5,6 +5,7 @@ import ReviewRecipientsPage from '../../pages/barcode/reviewRecipients'
 
 context('Create Barcode Image', () => {
   beforeEach(() => {
+    cy.task('reset')
```
mikehalmamoj, line 8: Doing the reset a in `goTo` method was restricting - couldn't set up a stub before calling that goTo method - better for each test to  perform the reset.

server/@types/express/index.d.ts
```diff
@@ -16,8 +16,8 @@ declare module 'express-session' {
     returnTo: string
     nowInMinutes: number
     requestLinkForm: RequestLinkForm
-    createBarcodeAuthToken: string
-    validCreateBarcodeAuthToken: boolean
+    slmToken: string
+    validSlmToken: boolean
```
mikehalmamoj, line 20: creatBarcodeAuthToken is being used for things other than creating barcodes (e.g. saving contacts - so renmaed this to the send legal mail token i.e. the token generated by send legal mail, not HMPPS Auth)

server/data/restClient.ts
```diff
@@ -87,85 +90,32 @@ export default class RestClient {
...
+      .timeout(this.timeoutConfig())
+
+    if (this.hmppsToken) {
+      request.auth(this.hmppsToken, { type: 'bearer' })
+    } else if (this.slmToken) {
+      request.set('Create-Barcode-Token', this.slmToken)
     }
-  }
 
-  async stream({ path = null, headers = {} }: StreamRequest = {}): Promise<unknown> {
```
mikehalmamoj, line 143: We don't stream anything - images are created locally

server/data/restClient.ts
```diff
@@ -23,19 +21,14 @@ interface PostRequest {
...
-
 export default class RestClient {
   agent: Agent
 
   constructor(
     private readonly name: string,
     private readonly config: ApiConfig,
-    private readonly token: string = undefined
+    private readonly hmppsToken: string = undefined,
+    private readonly slmToken: string = undefined
```
mikehalmamoj, line 31: Tried to distinguish between the HMPPS Auth token and Send Legal Mail generated token by their source of origin - they are both "auth" tokens, just different types.

server/data/restClient.ts
```diff
@@ -87,85 +90,32 @@ export default class RestClient {
...
-      return raw ? result : result.body
-    } catch (error) {
-      const sanitisedError = sanitiseError(error)
-      logger.warn({ ...sanitisedError }, `Error calling ${this.name}, path: '${path}', verb: 'POST'`)
-      throw sanitisedError
-    }
-  }
+    this.logRequest('POST', path)
 
-  async postCreateBarcode({
```
mikehalmamoj, line 113: Rolled this into the generic `post` method.

integration_tests/integration/barcode/createNewContactByPrisonNumber.spec.ts
```diff
@@ -4,6 +4,10 @@ import ReviewRecipientsPage from '../../pages/barcode/reviewRecipients'
 import CreateNewContactByPrisonNumberPage from '../../pages/barcode/createNewContactByPrisonNumber'
 
 context('Create New Contact By Prison Number Page', () => {
+  beforeEach(() => {
```
nathanrussell-moj-digital, line 7: Defo one for another day, but one of the reasons we moved all the `cy....` type gubbins out of the test was to abstract the test away from the browser driver, and let the `Page` subclass interact with the browser driver.
That knowledge is now sneaking back into the tests; so I wonder if (some other time) we look at other approaches for this type of thing? Maybe even something simple like creating a superclass for all cypress test specs called `CypressTest` or similar that has this `beforeEach` hook? If nothing else that would remove the duplication?
Just a thought 🤷‍♂️

integration_tests/mockApis/sendLegalMail.ts
```diff
@@ -306,6 +306,33 @@ const stubCreateBarcodeFailure = (): SuperAgentRequest =>
...
+      urlPattern: '/send-legal-mail/contact',
+      headers: {
+        'Create-Barcode-Token': {
+          equalTo:
+            'eyJhbGciOiJSUzI1NiJ9.eyJqdGkiOiJkZTE3NGM0Yi0xY2M2LTQxYWYtYTczYi01ZTE2YmI5YzE1ZWIiLCJzdWIiOiJtaWtlLmhhbG1hQGRpZ2l0YWwuanVzdGljZS5nb3YudWsiLCJleHAiOjQ3ODc1Njk3MjF9.WTqNajHRgZCbNe0g20lK5a7s_5-VeWD-FViu6gTgQaEsavimH_wEz1wZ4sj5osCDkCaLIgjYxGFt_p2IAsr7x0pI5b3CenN4_EMrz2pVVxAXOEEI8Q8QVfTy-iBGyO9W95rFGtmxbdsmYpr7LIr6DxJDUCCrCoeH8f4Dl-4QfKLUn-x_9_Bfum1rtAJ38B5pwiwhlzxeHD58C5XIc7swURGpCA97gtog7kEbyrCDF5AkIM4oYC1ViTMfDypnIJaDAU2ggxkaV5EkiIOB386POjUXkePQDnPajX3C-ugbJlKUPHp9z0CL_ngw5iK3wf9mEy2mWi9VHbUnyqVzfhrbIJK2PKQ0Fb8ZJIZhlB_rD68bgpaKskJwGy3lCMqDV5hiK5rUMsw_6n0asdYIhOvrEkXHrwmR4eRfobkLmtXGGRBswWuMhVXbYxBfZPU4PSkReTnbGRxSub-_UmMIvI_CXXaMdyRv0ixG4R3R7HfgLyZiTffN0p8nKmzKDXWWmPVJ',
+        },
+      },
+    },
+    response: {
+      status: 200,
```
nathanrussell-moj-digital, line 322: Minor, but the API returns a 201 for a successful creation

server/data/restClient.ts
```diff
@@ -48,17 +41,25 @@ export default class RestClient {
     return this.config.timeout
   }
 
-  async get({ path = null, query = '', headers = {}, responseType = '', raw = false }: GetRequest): Promise<unknown> {
-    if (this.token) {
-      logger.info(`GET request using authentication token: calling ${this.name}: ${path} ${query}`)
+  private logRequest(method: string, path: string, query?: string) {
+    if (this.hmppsToken) {
+      logger.info(
+        `${method.toUpperCase()} request using HMPPS auth token: calling ${this.name}: ${path} ${query ?? ''}`
```
nathanrussell-moj-digital, line 47: Nice refactoring of all this logging stuff 👍 
Out of interest, what is the double question mark syntax? I've not seen that before:
```
`${method.toUpperCase()} request using HMPPS auth token: calling ${this.name}: ${path} ${query ?? ''}`
```

integration_tests/integration/barcode/createNewContactByPrisonNumber.spec.ts
mikehalmamoj: Ah ok. It was a bit of a knee jerk reaction on my behalf because I created a stub which was then cleared by the goTo - took me ages to find the issue so jumped into the quickest fix. Yes let's have another look when we have a bit more time.

server/data/restClient.ts
mikehalmamoj: It's just yet another syntax for a null check - if null then do that. Thinking about it I could maybe have used a parameter default instead. Might change that in another PR...

PR #142, opened 2022-02-03
SLM-62 reafactor recipients

mikehalmamoj:
* Use a hidden RecipientForm to collect all recipient details between pages
* Hide all pages from each other
* Add a service for common RecipientForm functionality
* Refactor everything to use the new RecipientFormService
* And some readme updates

### DIFF
--- a/server/@types/forms/index.d.ts
+++ b/server/@types/forms/index.d.ts
@@ -1,5 +1,7 @@
 declare module 'forms' {
   import type { CheckBarcodeErrorCodes } from '../sendLegalMailApiClient'
+  import { PrisonAddress } from '../prisonTypes'
+  import { Contact } from '../sendLegalMailApiClient'
 
   export interface RequestLinkForm {
     email?: string
@@ -11,17 +13,14 @@ declare module 'forms' {
 
   export interface FindRecipientByPrisonerNameForm {
     prisonerName?: string
-    contacts?: Array<Contact>
   }
 
   export interface CreateNewContactByPrisonNumberForm {
-    prisonNumber: string
     prisonerName?: string
     prisonId?: string
   }
 
   export interface CreateNewContactByPrisonerNameForm {
-    prisonerName: string
     prisonId?: string
     prisonerDob?: Date
     'prisonerDob-day'?: string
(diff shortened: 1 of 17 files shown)


server/@types/forms/index.d.ts
```diff
@@ -11,17 +13,14 @@ declare module 'forms' {
 
   export interface FindRecipientByPrisonerNameForm {
     prisonerName?: string
-    contacts?: Array<Contact>
   }
 
   export interface CreateNewContactByPrisonNumberForm {
```
nathanrussell-moj-digital, line 18: I dont have an obvious answer to this, and I'm certainly not suggesting more refactoring as part of this, but I find the classnames of these forms (and their controllers) a bit misleading, especially now we've removed fields from them.
EG: The form class is called `CreateNewContactByPrisonNumberForm` .... the pertinent part being "By Prison Number" ... yet the form doesnt include a prison number field

As mentioned I dont have any suggestions at this point, but just thought I'd raise it

mikehalmamoj: Yes - I know what you mean. I'll have a think on that one...

server/routes/barcode/contacts/ChooseContactController.test.ts
```diff
@@ -34,24 +32,17 @@ const aContact = (): Contact => {
...
-
 describe('ChooseContactController', () => {
   let chooseContactController: ChooseContactController
 
   beforeEach(() => {
-    chooseContactController = new ChooseContactController(
-      contactService as unknown as ContactService,
-      prisonRegisterService as unknown as PrisonRegisterService
-    )
+    chooseContactController = new ChooseContactController(recipientFormService as unknown as RecipientFormService)
```
nathanrussell-moj-digital, line 39: Minor (and I'm not proposing you change it), but I think you can do this once without a beforeEach. Do we need to re-instantiate the controller between each test? Might be cleaner like this?:
```
describe('ChooseContactController', () => {
  const chooseContactController = new ChooseContactController(recipientFormService as unknown as RecipientFormService)
```

nathanrussell-moj-digital (approved):

PR #143, opened 2022-02-04
SLM-106: Add the Contact Helpdesk banner to every page

nathanrussell-moj-digital:
Screenshot 2022-02-04 at 09 37 31

### DIFF
--- a/integration_tests/pages/authManageDetails.ts
+++ b/integration_tests/pages/authManageDetails.ts
@@ -2,6 +2,6 @@ import Page from './page'
 
 export default class AuthManageDetailsPage extends Page {
   constructor() {
-    super('account-details', false)
+    super('account-details', { axeTest: false })
   }
 }
--- a/integration_tests/pages/page.ts
+++ b/integration_tests/pages/page.ts
@@ -5,9 +5,18 @@ export default abstract class Page {
     return new constructor()
   }
 
-  protected constructor(private readonly pageId: string, private readonly axeTest = true) {
+  protected constructor(
+    private readonly pageId: string,
+    private readonly options: { axeTest?: boolean; expectHelpdeskLink?: boolean } = {
+      axeTest: true,
+      expectHelpdeskLink: true,
+    }
+  ) {
     this.checkOnPage()
-    if (axeTest) {
+    if (options.expectHelpdeskLink) {
+      this.checkContactHelpdeskLink()
+    }
+    if (options.axeTest) {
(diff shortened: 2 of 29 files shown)


integration_tests/pages/authManageDetails.ts
```diff
@@ -2,6 +2,6 @@ import Page from './page'
 
 export default class AuthManageDetailsPage extends Page {
   constructor() {
-    super('account-details', false)
+    super('account-details', { axeTest: false })
```
nathanrussell-moj-digital, line 5: The `Page` constructor now has an options object as its 2nd arg - see `Page` below

integration_tests/pages/page.ts
```diff
@@ -5,9 +5,18 @@ export default abstract class Page {
     return new constructor()
   }
 
-  protected constructor(private readonly pageId: string, private readonly axeTest = true) {
+  protected constructor(
+    private readonly pageId: string,
+    private readonly options: { axeTest?: boolean; expectHelpdeskLink?: boolean } = {
+      axeTest: true,
+      expectHelpdeskLink: true,
```
nathanrussell-moj-digital, line 12: We expect the helpdesk link to be on every page, but in the event we get a requirement that says "Oh yeah, but not _this_ page" then we can handle that 👍

server/views/autherror.njk
```diff
@@ -3,10 +3,17 @@
...
+    <h1 class="govuk-heading-l">
+      Authorisation Error
+    </h1>
+    <p>
+      You are not authorised to use this application.
+    </p>
+
+  </section>
+
+  {{ contactHelpdeskBanner() }}
```
nathanrussell-moj-digital, line 18: Because of the way the govuk template works there is no obvious hook to enable us to inject content like this - it needs to sit just above the footer, but in respect of the markup it is within the main content.
The best I could do was to write a macro, and add it to every page just below the main `app-container` `section`
(there were a couple of pages like this one that didnt have a `section` to start with, so I've had to add it.

Basically, in respect of the markup, the new "contact the helpdesk" banner is a `section` that sits as a sibling to the main content `section`:

```
<section class="app-container govuk-body">
   ... all the main page content lives in here
</section>

<section class="contact-helpdesk">
    <a href="" class="govuk-link govuk-link--inverse">Contact the helpdesk</a>
</section>
```

mikehalmamoj (approved):

PR #145, opened 2022-02-04
SLM-121 Look for existing contact before creating a new one

mikehalmamoj:
Integration tests coming next (they still work only because our error handling ignores failing contact lookup!)


yasinmustafa (approved):
lgtm

PR #148, opened 2022-02-07
SLM-106: Initial form with pageId populated; cypress tests

nathanrussell-moj-digital:


integration_tests/pages/page.ts
```diff
@@ -34,9 +34,9 @@ export default abstract class Page {
     this.contactHelpdeskLink().should('exist').and('be.visible')
   }
 
-  contactHelpdesk = (): void => {
-    this.contactHelpdeskLink().click()
-    // TODO SLM-106 return next page once developed
+  contactHelpdesk<T>(T): T {
```
nathanrussell-moj-digital, line 37: For the life of me I could not get `Page` to refer to `ContactHelpdeskPage` (ie. our usual pattern of a method whose return type is the target page, and the impl does `return Page.verifyOnPage(ContactHelpdeskPage)`
It seems like its a circular dependency thing, but I just couldnt solve it, this is the best I could do :(

mikehalmamoj (approved):

PR #155, opened 2022-02-09
SLM-20 Add templates and controller for cookies policy

mikehalmamoj:


server/routes/cookies/CookiesPolicyController.ts
```diff
@@ -0,0 +1,53 @@
...
+      if (req.body.cookies === 'reject') {
+        Object.entries(req.cookies).forEach(cookie => {
+          if (typeof cookie[0] === 'string' && /^_g/.test(cookie[0])) {
+            res.clearCookie(cookie[0])
+          }
+        })
+      }
+      return res
+        .cookie('cookies_policy', req.body.cookies, { maxAge: 365 * 24 * 60 * 60 * 1000 })
+        .redirect(302, `${redirectUrl}?showCookieConfirmation=true`)
```
nathanrussell-moj-digital, line 45: Minor, but do we need the 302 in the `res.redirect` ? The only reason I ask is that we do `res.redirect` in all our other controllers and we dont send a 302 status code (I'd assumed up until this point that it did that for us ?)

mikehalmamoj: Ah probably not. I'll remove.

server/routes/cookies/CookiesPolicyController.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  async submitCookiesPolicyPreferences(req: Request, res: Response): Promise<void> {
+    let redirectUrl = req.session.cookiesPolicy.lastPage
+    if (req.originalUrl === '/cookies-policy') {
+      redirectUrl = '/cookies-policy'
+    }
+    if (req.body.cookies) {
+      if (req.body.cookies === 'reject') {
+        Object.entries(req.cookies).forEach(cookie => {
+          if (typeof cookie[0] === 'string' && /^_g/.test(cookie[0])) {
```
nathanrussell-moj-digital, line 38: Took me a while to grok this - is it basically saying "if you opted to reject cookies, we iterate through all the request cookies, and for each one, it element 0 (it's name ??) is a string and begins with _g (one of the google analytics cookies?), then we delete it.
Is that right?
I wonder if a comment, or refactoring into a method like "deleteGoogleAnaluyticsCookies" might help?

server/routes/cookies/CookiesPolicyController.ts
```diff
@@ -0,0 +1,53 @@
+import { NextFunction, Request, Response } from 'express'
+
+export default class CookiesPolicyController {
+  async initialiseCookieSession(req: Request, res: Response, next: NextFunction): Promise<void> {
```
nathanrussell-moj-digital, line 4: What does this method do and where is/will it be called from?
I can't quite grok why there is different behaviour for external vs. internal users ?

server/routes/cookies/CookiesPolicyController.ts
mikehalmamoj: Good call - I pinched that code from somewhere

server/routes/cookies/CookiesPolicyController.ts
mikehalmamoj: It's called in an app.use when setting up the cookie routes. It's basically there to save the existing cookie in the session and fill locals that can be used in views.

The difference between internal and external users is that internal users won't get the cookie banner or get a chance to opt out, as is standard across internal DPS apps.

nathanrussell-moj-digital (approved):

PR #157, opened 2022-02-10
SLM-20 secure cookies

mikehalmamoj:
* and update jQuery
* and serve source map (that broke the app if opening dev tools in the browser. Only on Chrome. 🤣  Thanks @nathanrussell-moj-digital!)

### DIFF
--- a/server/views/layout.njk
+++ b/server/views/layout.njk
@@ -37,10 +37,8 @@
   <![endif]-->
 
   <script src="/assets/js/jquery.min.js"></script> 
-  <script src="https://code.jquery.com/ui/1.12.1/jquery-ui.min.js"
-          integrity="sha256-VazP97ZCwtekAsvgPBSUwPFKdrwD3unUfSGVYrahUqU="
-          crossorigin="anonymous"></script>
-  <link href="https://code.jquery.com/ui/1.12.1/themes/ui-lightness/jquery-ui.css" rel="stylesheet">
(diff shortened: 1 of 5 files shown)


server/views/layout.njk
```diff
@@ -37,10 +37,8 @@
   <![endif]-->
 
   <script src="/assets/js/jquery.min.js"></script> 
-  <script src="https://code.jquery.com/ui/1.12.1/jquery-ui.min.js"
-          integrity="sha256-VazP97ZCwtekAsvgPBSUwPFKdrwD3unUfSGVYrahUqU="
-          crossorigin="anonymous"></script>
-  <link href="https://code.jquery.com/ui/1.12.1/themes/ui-lightness/jquery-ui.css" rel="stylesheet">
+  <script src="https://code.jquery.com/ui/1.13.1/jquery-ui.min.js" integrity="sha256-eTyxS0rkjpLEo16uXTS0uVCS4815lc40K2iVpWDvdSY=" crossorigin="anonymous"></script>
```
nathanrussell-moj-digital, line 40: Do we really want to be using jquery 1? I dont think we need IE 6 or 8 support do we? so we could use v2, and better still v3 ?

https://stackoverflow.com/questions/41605120/what-is-the-difference-between-jquery-version-1-version-2-and-version-3

I wonder if this is a question for the wider DPS devs community as to whether anyone else has moved from jquery 1 to a more recent/modern version ?

nathanrussell-moj-digital (approved):

PR #159, opened 2022-02-11
SLM-20 Use Google recommended way to opt out of GA

mikehalmamoj:


nathanrussell-moj-digital (approved):
That looks like a better way of doing it 👍

PR #161, opened 2022-02-14
Fix parsing of booleans in config

nathanrussell-moj-digital:


server/config.ts
```diff
@@ -2,12 +2,19 @@ import 'dotenv/config'
 
 const production = process.env.NODE_ENV === 'production'
 
-function get<T>(name: string, fallback: T, options = { requireInProduction: false }): T | string {
+const toBoolean = (value: unknown): boolean => {
+  return value === 'true'
+}
+
+function get(name: string, fallback: unknown, options = { requireInProduction: false }): string | null {
```
nathanrussell-moj-digital, line 9: The original `get` method was the root cause of the problem. It's original signature suggested that it would return a type, when in reality it would only return a type via the `fallback` only when the corresponding env variable was not found. If the env variable was found it would return a `string` because env vars can only be string.

eg:
* `get('FILE_UPLOADS_ENABLED', false)` returned `false` if `FILE_UPLOADS_ENABLED` is not setup in the environment
* `get('FILE_UPLOADS_ENABLED', false)` returned `"false"` if `FILE_UPLOADS_ENABLED` is setup in the environment as "false"
* `get('FILE_UPLOADS_ENABLED', false)` returned `"true"` if `FILE_UPLOADS_ENABLED` is setup in the environment as "true"

The last 2 use cases are problematic because `"false" !== false`, and the `Boolean`constructor does not help! - `new Boolean("false") === true` !!

mikehalmamoj (approved):

PR #169, opened 2022-02-15
SLM-141: Updated CSP header

nathanrussell-moj-digital:


server/middleware/setUpWebSecurity.ts
```diff
@@ -19,19 +19,22 @@ export default function setUpWebSecurity(): Router {
       contentSecurityPolicy: {
         directives: {
           defaultSrc: ["'self'"],
-          // Hash allows inline script pulled in from https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/template.njk
```
nathanrussell-moj-digital, line 22: The comment about the hash, and the hash itself are outdated so I have removed them. The script in the template that it refers to has since been changed to use the nonce value, so it doent need this anymore (and the hash will be different anyway) https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/template.njk#L31

I've also re-ordered the values slightly as well - self, then the nonce value, then external hosts

server/middleware/setUpWebSecurity.ts
```diff
@@ -19,19 +19,22 @@ export default function setUpWebSecurity(): Router {
...
             'www.google-analytics.com',
-            (req: IncomingMessage, res: Response) => `'nonce-${res.locals.cspNonce}'`,
             'code.jquery.com',
-            "'sha256-+6WnXIl4mbFTCARd8N3COQmT3bJJmo32N8q8ZSQAIcU='",
           ],
           connectSrc: ["'self'", 'www.googletagmanager.com', 'www.google-analytics.com'],
           styleSrc: ["'self'", 'code.jquery.com'],
           fontSrc: ["'self'"],
           imgSrc: ["'self'", 'data:'],
+          objectSrc: ["'none'"],
```
nathanrussell-moj-digital, line 33: All of these were recommendations from https://csp-evaluator.withgoogle.com/

mikehalmamoj (approved):

PR #176, opened 2022-02-17
SLM-63 update contact api

mikehalmamoj:


server/data/restClient.ts
```diff
@@ -83,16 +84,17 @@ export default class RestClient {
       })
   }
 
-  async post({
+  async update({
```
mikehalmamoj, line 87: @nathanrussell-moj-digital - I wrote a put function that looked exactly the same as the post function so I merged them into one. I had no idea what to call this new function so I went with update. I don't like it. Do you have any better ideas?

nathanrussell-moj-digital: Given that this class is `restClient` I would say there is an assumed level of knowledge as to what someone is trying to do, so using 'rest' (HTTP) terms doesn't feel unreasonable to me. How about taking inspiration('ish) from Spring's `RestTemplate` as conceptually this class is not a million miles away from it?

How about 3 public methods - `get`, `post` and `put`. Given that the post and the put are so similar then I agree, you want to refactor in some way. And then a private method `exchange` (which is basically your proposed `update` method). Your 2 public `post` and `put` methods could then just delegate to the private `exchange` method, passing in the required HTTP verb.

Would that work?

server/services/helpdesk/ZendeskService.ts
```diff
@@ -44,7 +44,7 @@ Email: ${contactHelpdeskForm.email}
           tags: externalUser ? 'slm_legal_sender' : 'slm_mailroom',
         },
       }
-      const response = (await ZendeskService.restClient().post({
+      const response = (await ZendeskService.restClient().update({
```
nathanrussell-moj-digital, line 47: Woah, hang on, why have these changed to `.update`? This is not an update operation, its a POST for a new entity

server/services/link/MagicLinkService.ts
```diff
@@ -13,7 +13,7 @@ export default class MagicLinkService {
   async requestLink(email: string): Promise<unknown> {
     const magicLinkRequest: MagicLinkRequest = { email }
     return this.hmppsAuthClient.getSystemClientToken().then(hmppsToken =>
-      MagicLinkService.restClient(hmppsToken).post({
+      MagicLinkService.restClient(hmppsToken).update({
```
nathanrussell-moj-digital, line 16: Likewise - this is not an update operation.
I can now see why you were looking to explore other ways of refactoring the `restClient` class. As mentioned in my comment on it, I think the public API of the rest client should be rest/http terminology, because otherwise we end up with confusion like this at the service layer (ie. I've read this line and am immediately assuming I'm updating an existing magic link entity/resource, which is not the case)

server/data/restClient.ts
```diff
@@ -90,31 +73,59 @@ export default class RestClient {
...
+    }
+    if (query) {
+      request.query(query)
+    }
+
     if (this.hmppsToken) {
       request.auth(this.hmppsToken, { type: 'bearer' })
     } else if (this.slmToken) {
       request.set('Create-Barcode-Token', this.slmToken)
     }
```
nathanrussell-moj-digital, line 122: I like this a lot (a single method to invoke the actual http request), and right here is where I'm going to add basic auth support in my next PR 👍 😄

nathanrussell-moj-digital (approved):
LGTM

PR #182, opened 2022-02-21
SLM-128: Fixed PDF Download link by copying recipients into a new array

nathanrussell-moj-digital:


assets/js/page-enhancements.js
```diff
@@ -82,7 +82,9 @@
 })($, document)
 ;(autoDownloadCoversheetPdf = ($, document) => {
   $(document).ready(() => {
-    $('#print-coversheets-content #downloadPdf').each((idx, element) => element.click())
+    setTimeout(() => {
```
nathanrussell-moj-digital, line 85: Delaying the auto-download on page render by 1 second is better as it allows the page to fully render and gives cypress a fighting chance of being able to assert the page properly. And 1 second doesnt impact the UX at all

integration_tests/pages/barcode/printCoversheets.ts
```diff
@@ -1,16 +1,16 @@
 /* eslint-disable import/no-cycle */
 import Page from '../page'
 import SelectEnvelopeSizePage from './selectEnvelopeSize'
+import FindRecipientByPrisonNumberPage from './findRecipientByPrisonNumber'
 
 export default class PrintCoversheetsPage extends Page {
   constructor() {
-    super('print-coversheets', { axeTest: false })
```
nathanrussell-moj-digital, line 7: Now there is a 1 second delay before JS tries to auto-download the PDF we can remove this config and comment and allow nice things like axe testing to work 👍

integration_tests/integration/barcode/pdfCoversheet.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+context('Print PDF coversheet', () => {
+  let printCoversheetsPage: PrintCoversheetsPage
+
+  beforeEach(() => {
+    cy.task('reset')
+    cy.task('stubCreateBarcode')
+    printCoversheetsPage = PrintCoversheetsPage.goToPage()
+  })
+
+  it('should allow the PDF to be downloaded again', () => {
```
nathanrussell-moj-digital, line 14: I was desperately trying to write methods in the page class to assert the file had been downloaded, but 'varying degrees of success' is the polite way of saying how far I got with it!
The code was getting increasingly hacky and indeterminate, and in the end I decided it wasnt worth the effort

integration_tests/pages/barcode/printCoversheets.ts
```diff
@@ -1,17 +1,26 @@
...
-     axeTest (2nd constructor arg) must be false because axe relies on the browser tab being the active tab so that it can
-     interrogate the DOM. This page however auto-downloads the coversheet PDF as part of page load, so by the time axe runs
-     the active tab is not the page under test (even though the tab is still visible and the PDF has been downloaded rather than
-     opened in a new tab)
-     */
+    super('print-coversheets')
   }
 
+  downloadThePdfAgain = (): PrintCoversheetsPage => {
+    this.downloadLink().then(link => {
```
nathanrussell-moj-digital, line 12: This is one of the hacks I mentioned above that I've had to keep. The obvious thing to have done is:
```
this.downloadLink().click()
```
But cypress expects a `click` event to render a new page and so waits for a page load event from the browser ... which in this case never comes!
So we have to use cypress' `then` method to treat the element as a jquery element and use jquery's API to click it! 😢

mikehalmamoj (approved):

PR #191, opened 2022-02-23
SLM-135: Added clamav to docker-compose, helm, and circle

nathanrussell-moj-digital:


.circleci/config.yml
```diff
@@ -17,6 +17,14 @@ parameters:
     type: string
     default: 14.17-browsers
 
+executors:
+  integration-test:
```
nathanrussell-moj-digital, line 21: This basically copys what is in the named executor `hmpps/node_redis` (starting a node and redis container), then also starts the clamav container.
I couldnt work out how 'extend' an executor, or use 2 executors for a job ... 🤷‍♂️

mikehalmamoj: No no-one does! I asked about this the other week! I think we need some way of having a single executor where you can select which images to include... but the godfather of orbs left a while back.

mikehalmamoj (approved):

PR #193, opened 2022-02-23
SLM-63 update contact

mikehalmamoj:


server/routes/barcode/recipients/RecipientFormService.ts
```diff
@@ -37,24 +37,47 @@ export default class RecipientFormService {
...
       req.session.recipients = []
     }
+    const newRecipient = await this.toRecipient(recipientForm)
+
+    req.session.recipients.push(newRecipient)
+
+    req.session.recipientForm = {}
+  }
+
+  private async updateRecipient(req: Request, recipientForm: RecipientForm) {
```
nathanrussell-moj-digital, line 47: Does this method need to be async? Its not calling any external service is it? Its just updating the array of recipients in the session with the updated contact/recipient isnt it? 
(having said that it looks like the other methods in this class are also async, so perhaps theres no harm?)

server/middleware/barcode/setupBarcode.ts
```diff
@@ -89,6 +89,7 @@ export default function setUpCreateBarcode(
   router.get('/generate-barcode-image', (req, res) => generateImageController.getGenerateImageView(req, res))
 
   router.get('/edit-contact/:contactId', (req, res) => editContactController.getEditContact(req, res))
+  router.post('/edit-contact', (req, res) => editContactController.submitUpdateContact(req, res))
```
nathanrussell-moj-digital, line 92: Did you consider making this POST endpoint `/edit-contact/:contactId` like the GET endpoint? A bit more restful (though of course its a web app, so is not rest at all 😁  )

server/routes/barcode/recipients/RecipientFormService.ts
mikehalmamoj: Yes it ends up calling prison register service to get the prison address - which is async

PR #197, opened 2022-02-25
SLM-63 Wire up the edit contact link and add integration tests

mikehalmamoj:


integration_tests/mockApis/sendLegalMail.ts
```diff
@@ -351,11 +351,10 @@ const stubCreateContact = (): SuperAgentRequest =>
         'Content-Type': 'application/json;charset=UTF-8',
       },
       jsonBody: {
-        id: 1,
-        prisonerName: `{{jsonPath request.body '$.prisonerName'}}`,
-        prisonId: `{{jsonPath request.body '$.prisonId'}}`,
-        prisonNumber: `{{jsonPath request.body '$.prisonNumber'}}`,
-        prisonerDob: `{{jsonPath request.body '$.prisonerDob'}}`,
```
mikehalmamoj, line 358: This wasn't working! 😨 And no amount of hacking has fixed, so hardcoding required.

server/filters/recipientTableRowsFilter.ts
```diff
@@ -14,7 +14,7 @@ export default function recipientTableRowsFilter(recipients: Array<Recipient>):
         text: recipient.prisonNumber ? recipient.prisonNumber : moment(recipient.prisonerDob).format('DD-MM-YYYY'), // Column 2 - Prison number or DOB
       },
       { text: recipient.prisonAddress.premise }, // Column 3 - Prison name
-      { html: '<a href="" class="govuk-link">Edit details</a>' }, // Column 4 - Edit recipient details link - TODO SLM-63
+      { html: `<a href="/barcode/edit-contact/${recipient.contactId}" class="govuk-link">Edit details</a>` }, // Column 4 - Edit recipient details link
       { html: `<a href="/barcode/review-recipients/remove/${idx}" class="govuk-link">Remove</a>` } // Column 5 - Remove recipient link
```
nathanrussell-moj-digital, line 18: Not one for now, but perhaps we should change the 'Remove' anchor and its corresponding controller method to take the contact ID rather than the index in the array. Removing by contactID feels much safer than removing by index from an array (even though its only removing from an in-memory array structure rather than making a rest call like Edit is) 
Just a thought

mikehalmamoj: Ah yes - sounds good

integration_tests/mockApis/sendLegalMail.ts
```diff
@@ -333,7 +333,7 @@ const stubCreateBarcodeFailure = (prisonerName?: string): SuperAgentRequest => {
   })
 }
 
-const stubCreateContact = (): SuperAgentRequest =>
+const stubCreateContact = (contactId: number): SuperAgentRequest =>
```
nathanrussell-moj-digital, line 336: I was going to ask this, but the penny has dropped! 😁 

I _was_ going to ask why we've changed the create contact endpoint to take the contact ID, and to question how that would work because we wouldn't know the contact ID at that point, but then the penny dropped and I realised that this was the stub and that the contactID arg is for building the response! 👍  

Doh! 😁

nathanrussell-moj-digital (approved):

PR #208, opened 2022-03-02
SLM-151: Cookie preferences can now be submitted as ajax requests to reduce duplicate GA page impressions

nathanrussell-moj-digital:


assets/js/page-enhancements.js
```diff
@@ -7,9 +7,8 @@
  *
  * No code should be written here that renders a page non-functional if JS is not enabled or the script does not run/fails.
  */
-
```
nathanrussell-moj-digital, line 10: I've changed how the js page enhancements stuff works slighty.
Previously we had a load of named iife's, where each one handled a specific page enhancement. As function expressions, they were 'isolated' in that they could not reference other things outside of their own scope.
This meant that we couldnt have common code in a function that was called by the iife's (unless we defined the function globally with is a bit meh), and also meant one iife could not call another iife. Both of these scenarios turned out to be something we now want to do :) 

What I've done is use the revealing module pattern. Its a iife that internally defines a load of functions - the same ones that we previously had as individual ifee's. But they are effectively private because they are within the scope of the function.
The function returns an object that exposes anything we want to expose. This pattern is often used to expose functions that we want to be public. Its basically a poor man's module before modules were a proper thing, but it serves our purpose well here. In our returned object we are returning a function `init` which when called will call each of the private methods - ie. exactly what all the previous individual iife's did.
The benefit here is that we can have common code as private functions that the other functions can call (see `extractCsrfFieldAsKeyValuePairFromForm`), and also one function can call another function to effectively run it again (see the ajax call back in `submitCookiePreferencesFormViaAjax` that calls `submitCookiePreferenceConfirmationFormViaAjax` to apply page enhancements to the returned page partial)

assets/js/page-enhancements.js
```diff
@@ -80,17 +78,77 @@
...
+      .map(formField => {
+        return `${formField.name}=${formField.value}`
+      })
+  }
+
+  const submitCookiePreferencesFormViaAjax = () => {
+    $('#cookiePreferencesForm button[type=submit]').on('click', function (event) {
+      event.preventDefault()
+      const form = $(this).closest('form')
+      const csrfQueryString = extractCsrfFieldAsKeyValuePairFromForm(form)
```
nathanrussell-moj-digital, line 106: Because its an ajax request (appliction/json), traditional form fields cannot be sent in the same way as a browser POST request does. In respect of the `_csrf` field we need to extract it and add it as a query string parameter. All other fields/data (if there are any) can be serialized into an object and sent as the request body

PR #222, opened 2022-03-08
SLM-85: Added tests using supertest that spin up and test the express application

nathanrussell-moj-digital:


integration_tests/pages/page.ts
```diff
@@ -23,7 +23,7 @@ export default abstract class Page {
   }
 
   checkOnPage = (): void => {
-    cy.get(`#${this.pageId}`).should('exist')
+    cy.get('#pageId').should('have.attr', 'data-qa').should('equal', this.pageId)
   }
```
nathanrussell-moj-digital, line 27: The span with the pageId has changed in the template - see below 👇

package.json
```diff
@@ -63,7 +63,8 @@
       "json",
       "node",
       "ts"
-    ]
+    ],
+    "testRunner": "jest-jasmine2"
```
nathanrussell-moj-digital, line 67: We need to include the jasmine runner to be able to use `fail` in tests - its basically a bug in `jest-circus` - see https://github.com/facebook/jest/issues/11698

server/routes/barcode/pdf/selectEnvelopeSize.spec.ts
```diff
@@ -0,0 +1,28 @@
+import SuperTestWrapper from '../../fixtures/SuperTestWrapper'
+import assertThat from '../../fixtures/supertest-assertions'
+
+describe('Select envelope size', () => {
+  const superTest = new SuperTestWrapper()
+
+  beforeEach(async () => {
```
nathanrussell-moj-digital, line 7: I've added a test that requires the user to be authenticated. The test spec itself could do with some more functional tests adding (ie. submitting the envelope size etc), but I wanted to make sure we could send requests as an authenticated user

server/routes/fixtures/SuperTestWrapper.ts
```diff
@@ -0,0 +1,37 @@
...
+import request from 'supertest'
+// eslint-disable-next-line import/no-extraneous-dependencies
+import nock from 'nock'
+import express from 'express'
+import app from '../../index'
+import config from '../../config'
+
+jest.mock('redis', () => jest.requireActual('redis-mock'))
+
+export default class SuperTestWrapper {
```
nathanrussell-moj-digital, line 11: This is the main class you are going to use when writing supertest tests - basically new up a `SuperTestWrapper` and use it's `request` property to issue requests.
It also has helper methods to make the test context authenticated or unauthenticated (need to add a method to authenticate as a mail room user, but that will come ....)

server/routes/fixtures/SuperTestWrapper.ts
```diff
@@ -0,0 +1,37 @@
...
+      token:
+        'eyJhbGciOiJSUzI1NiJ9.eyJqdGkiOiJkZTE3NGM0Yi0xY2M2LTQxYWYtYTczYi01ZTE2YmI5YzE1ZWIiLCJzdWIiOiJtaWtlLmhhbG1hQGRpZ2l0YWwuanVzdGljZS5nb3YudWsiLCJleHAiOjQ3ODc1Njk3MjF9.WTqNajHRgZCbNe0g20lK5a7s_5-VeWD-FViu6gTgQaEsavimH_wEz1wZ4sj5osCDkCaLIgjYxGFt_p2IAsr7x0pI5b3CenN4_EMrz2pVVxAXOEEI8Q8QVfTy-iBGyO9W95rFGtmxbdsmYpr7LIr6DxJDUCCrCoeH8f4Dl-4QfKLUn-x_9_Bfum1rtAJ38B5pwiwhlzxeHD58C5XIc7swURGpCA97gtog7kEbyrCDF5AkIM4oYC1ViTMfDypnIJaDAU2ggxkaV5EkiIOB386POjUXkePQDnPajX3C-ugbJlKUPHp9z0CL_ngw5iK3wf9mEy2mWi9VHbUnyqVzfhrbIJK2PKQ0Fb8ZJIZhlB_rD68bgpaKskJwGy3lCMqDV5hiK5rUMsw_6n0asdYIhOvrEkXHrwmR4eRfobkLmtXGGRBswWuMhVXbYxBfZPU4PSkReTnbGRxSub-_UmMIvI_CXXaMdyRv0ixG4R3R7HfgLyZiTffN0p8nKmzKDXWWmPVJ',
+    })
+
+    await this.request.get('/link/verify-link?secret=some-secret')
+    nock.cleanAll()
+  }
+
+  unauthenticated = async () => {
+    await this.request.get('/link/request-link?force=true') // log out legal sender user
```
nathanrussell-moj-digital, line 34: The easiest way to be unauthenticated is to hit the 2 sign out links - it doesnt matter whether you were previously logged in as a LS or MR user, or neither.

server/routes/fixtures/supertest-assertions.ts
```diff
@@ -0,0 +1,69 @@
+import * as superagent from 'superagent'
+// eslint-disable-next-line import/no-extraneous-dependencies
+import cheerio, { CheerioAPI } from 'cheerio'
+
+export default function assertThat(response: superagent.Response) {
+  return new SupertestAssertions(response)
+}
+
+class SupertestAssertions {
```
nathanrussell-moj-digital, line 9: This is effectively our DSL for making assertions. All methods are chainable and is modelled heavily on assertJ (mainly because I couldnt work out how to write my own `expect` method that didnt conflict with jest's `expect`)

Also, in most assertion methods I have a try/catch with a `fail` (re: the change to package.json to bring in jasmine)
Its a bit long winded but its because jest's `expect` method doesnt support failure messages 😢 
Long winded? yes; but its nicely abstracted away into this class, so test methods dont need to worry about it

server/routes/fixtures/supertest-assertions.ts
```diff
@@ -0,0 +1,69 @@
...
+    return this
+  }
+
+  hasError = (expected: string): SupertestAssertions => {
+    this.hasAllErrors(expected)
+    return this
+  }
+
+  hasAllErrors = (...expected: string[]): SupertestAssertions => {
+    const actual = this.$('.govuk-error-summary__list').text()
```
mikehalmamoj, line 52: Minor point - might be better to split this by newlines before testing? Meh.

server/routes/link/VerifyLinkController.ts
```diff
@@ -35,8 +35,9 @@ export default class VerifyLinkController {
         req.flash('errors', [
           { href: '#email', text: 'The link you used is no longer valid. Request a new one to sign in.' },
         ])
-        res.redirect('/link/request-link')
+        return res.redirect('/link/request-link')
```
mikehalmamoj, line 38: do we also need this on line 28? bit confused as to how this ever worked!

server/routes/link/requestLink.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+    mockHmppsAuth()
+    mockedSendLegalMailApi = nock(config.apis.sendLegalMail.url)
+  })
+
+  afterEach(() => {
+    nock.cleanAll()
+  })
+
+  it('should render page to request a link', async () => {
+    const response = await superTest.request //
```
mikehalmamoj, line 22: Does the `////` at the end mean anything?

nathanrussell-moj-digital: No, it's just a way I can shoehorn my desire to lay out the code into a more readable manner (IMHO) rather than being at the whim of the draconian prettier rules 😁 
To be fair this particular instance is not that offensive because there is only one method call being chained:
```
const response = await superTest.request //
      .get('/link/request-link')
```
Without the comment markers prettier would force me to do:
```
const response = await superTest.request.get('/link/request-link')
```

That's no _so_ bad, but there are other cases below where I chain more method calls:
```
const response = await superTest.request //
      .post('/link/request-link')
      .send({ email: 'user@aardvark.com.cjsm.net' })
```
If prettier had it's way it would force me to do:
```
const response = await superTest.request.post('/link/request-link').send({ email: 'user@aardvark.com.cjsm.net' })
```
Horrible 🤮

server/routes/link/VerifyLinkController.ts
nathanrussell-moj-digital: I'll explain it tomorrow, but basically it worked by luck by virtue of the fact that the error message rendered in the flash object for both use cases is the same (lines 26 and 36).
If they had different error messages we would have noticed this bug sooner.
The only reason I noticed it here is that `res.render` was being called twice (I'll explain why tomorrow) and the `supertest` http client (`superagent`) complained bitterly that the response had already been written to and that it couldnt do it again!

PR #231, opened 2022-03-15
SLM-190 post not get

mikehalmamoj:


server/data/cache/SmokeTestStore.ts
```diff
@@ -0,0 +1,35 @@
...
+    return this.setRedisAsync(SMOKE_TEST, oneTimeSecret, 'EX', 60)
+  }
+
+  public async getSmokeTestSecret(): Promise<string> {
+    const secret = this.getRedisAsync(SMOKE_TEST)
+    this.deleteEntry(SMOKE_TEST)
+    return secret
+  }
+
+  public async startSmokeTest(req: Request): Promise<string> {
```
nathanrussell-moj-digital, line 25: Minor, and we can save it for another day, but I'm not sure about this redis store having knowledge of express and its `Request` type.
I wonder if the method signature could just take the msjSecret as a string, and the thing thats calling this be responsible for getting it out of the request body?
One for another day ... 👍

server/middleware/smoketest/SmokeTestMiddleware.ts
```diff
@@ -0,0 +1,10 @@
+import express, { Router } from 'express'
+import SmokeTestStore from '../../data/cache/SmokeTestStore'
+
+export default function setupSmokeTest(smokeTestStore: SmokeTestStore): Router {
+  const router = express.Router()
+
+  router.post('/getSmokeTestSecret', async (req, res) => res.json({ token: await smokeTestStore.startSmokeTest(req) }))
```
nathanrussell-moj-digital, line 7: Re: my comment above .... ah, its called here .... this code could look quite complex if we implemented what I suggested ... Lets leave it for now

nathanrussell-moj-digital (approved):

PR #234, opened 2022-03-19
SLM-190 Support smoke test

mikehalmamoj:
Add barcopde number to screen when running the LSJ smoketest so it can be used in the LSJ smoketest.

### DIFF
--- a/server/views/pages/barcode/pdf/print-coversheets.njk
+++ b/server/views/pages/barcode/pdf/print-coversheets.njk
@@ -59,6 +59,12 @@
         <a href="/barcode/find-recipient" data-qa="send-more-legal-mail" class="govuk-link">Enter another prison number</a>
       </div>
     </div>
+
+    {% if smokeTestBarcode %}
+    <div class="govuk-visually-hidden">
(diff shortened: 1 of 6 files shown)


server/views/pages/barcode/pdf/print-coversheets.njk
```diff
@@ -59,6 +59,12 @@
         <a href="/barcode/find-recipient" data-qa="send-more-legal-mail" class="govuk-link">Enter another prison number</a>
       </div>
     </div>
+
+    {% if smokeTestBarcode %}
+    <div class="govuk-visually-hidden">
```
nathanrussell-moj-digital, line 64: Very minor because this block will never be rendered in production, but an element with class `govuk-visually-hidden` will be read out loud by a screen reader but not rendered visually.
Given that it's cypress that is obtaining the element programatically from the DOM I'd have possibly gone with `display:none` (or a govuk equivalent)
But as mentioned its not ever going to be rendered in production so 🤷‍♂️ 😁

nathanrussell-moj-digital (approved):

PR #263, opened 2022-04-05
SLM-221: Added recipient name to copy and download buttons

nathanrussell-moj-digital:


server/routes/barcode/image/GenerateBarcodeImageView.ts
```diff
@@ -1,6 +1,7 @@
 type BarcodeImage = {
   barcodeImageUrl: string
   barcodeImageName: string
+  recipientName: string
```
nathanrussell-moj-digital, line 4: The recipient name needs to be added to the barcode image data in the view, so that the nunjucks template can render it as part of the Copy and Download buttons

mikehalmamoj (approved):

PR #267, opened 2022-04-06
SLM-202 additional GA data

mikehalmamoj:
I didn't want to add this to layout.njk as it doesn't apply to all pages - so I'm trying this method of pushing to GA using JS.

### DIFF
--- a/assets/js/page-enhancements.js
+++ b/assets/js/page-enhancements.js
@@ -144,6 +144,15 @@ window.pageEnhancements = (($, document) => {
     })
   }
 
+  const pushGtmBarcodeCount = () => {
+    const barcodeCreateCount = $('#barcode-created-count').attr('data-value')
+    if (barcodeCreateCount) {
+      window.dataLayer.push({
+        barcodesCreatedCount: barcodeCreateCount,
+      })
+    }
+  }
+
   return {
     init: () => {
       $(() => {
@@ -153,6 +162,7 @@ window.pageEnhancements = (($, document) => {
         autoFocusBarcodeField()
         submitCookiePreferencesFormViaAjax()
         submitCookiePreferenceConfirmationFormViaAjax()
+        pushGtmBarcodeCount()
       })
     },
   }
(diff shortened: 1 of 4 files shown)


server/views/pages/barcode/generate-barcode-image.njk
```diff
@@ -19,6 +19,7 @@
       }) }}
     {% endif %}
 
+    <span id="barcode-created-count" class="govuk-visually-hidden" data-value="{{ barcodeImages | length }}"></span>
```
nathanrussell-moj-digital, line 22: I'm not sure visually hidden is the right class is it? I _think_ that whilst it wont be visible on screen, a screen reader will still read out the element .... so a visually impaired person will hear a random number as part of the reading of the page! 😁 

I think what you want is `display: none`, and I think there is a convenient govuk style you can use. I think its `govuk-!-display-none` (but you might want to check that)

assets/js/page-enhancements.js
```diff
@@ -144,6 +144,23 @@ window.pageEnhancements = (($, document) => {
...
 
+  const pushGtmBarcodeCount = () => {
+    const barcodesCreatedCount = $('#barcode-created-count').attr('data-value')
+    if (barcodesCreatedCount) {
+      window.dataLayer.push({
+        barcodesCreatedCount: barcodesCreatedCount,
+      })
+    }
+  }
+  const pushGtmScannedBarcodeNumber = () => {
```
nathanrussell-moj-digital, line 155: I know its part of the ACs, but I'm wondering what the value of pushing the scanned barcode number to GA is? What will they do with that info?
I'm also slightly uncomfortable about sending the barcode number, though I cant qualify it. I know the number will have been scanned at that point so cannot be reused; but something is not quite sitting right with me. Maybe understanding the use case as to why they want the scanned barcode number might alleviate my concerns ?

nathanrussell-moj-digital (approved):

PR #268, opened 2022-04-07
SLM-202 fix ga timing

mikehalmamoj:
Using JS to push to Google Analytics fails - seems to be sending requests to gtm out of sync.

Revert back to the proven method - push to gtm before gtm consent is initialised

### DIFF
--- a/server/views/layout.njk
+++ b/server/views/layout.njk
@@ -34,6 +34,14 @@
         });
     {% endif %}
 
+    {% if numberOfCoversheets %}
+      window.dataLayer.push({
+          'user': {
+            'barcodes_created_count': '{{ numberOfCoversheets }}',
+          },
+        });
+    {% endif %}
+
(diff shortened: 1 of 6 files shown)


server/views/layout.njk
```diff
@@ -26,6 +26,22 @@
         });
     {% endif %}
 
+    {% if barcodeImages | length %}
```
nathanrussell-moj-digital, line 29: Do you need a similar block of code for `numberOfCoversheets` ?

nathanrussell-moj-digital: Or does the number of barcodeImages cover both copy/paste barcode & PDF journeys?

mikehalmamoj: Oh yeah! Thanks.

nathanrussell-moj-digital (approved):

PR #273, opened 2022-04-11
SLM-231: Imported types from OpenAPI docs; added logic to use correct LSJ start page

nathanrussell-moj-digital:
PR to import the updated types from the API's openAPI docs (new types for the new One Time Code endpoints); added a helper function to correctly return the correct start page URL


server/@types/sendLegalMailApi/index.d.ts
```diff
@@ -71,6 +79,19 @@ export interface components {
       code: string
       userMessage: string
     }
+    AuthenticationRequestErrorCodes: components['schemas']['ErrorCode'] & {
```
nathanrussell-moj-digital, line 82: This is the new name for the type in the API (it was previously called `MagicLinkRequestErrorCodes`)

server/middleware/legalSenderJourneyAuthenticationStartPage.ts
```diff
@@ -0,0 +1,5 @@
+import config from '../config'
+
+export default function legalSenderJourneyAuthenticationStartPage() {
```
nathanrussell-moj-digital, line 3: Function to correctly return the correct start page URL for the Legal Sender Journey based on the value of the feature flag (ie. either use magic link or One Time Code urls/routes) - means we dont need to replicate the logic in several places

joelstobart-moj (approved):
LGTM

PR #293, opened 2022-04-22
SLM-195: Update prison register types; change PrisonRegisterStore to use PrisonDto

nathanrussell-moj-digital:


server/@types/prisonRegisterApi/index.d.ts
```diff
@@ -48,6 +127,80 @@ export interface components {
...
+      town: string
+      /** County */
+      county?: string
+      /** Postcode */
+      postcode: string
+      /** Country */
+      country: string
+    }
+    /** Prison Information */
+    PrisonDto: {
```
nathanrussell-moj-digital, line 148: These are the types we are _really_ interested in - `PrisonDto` and `AddressDto`

server/data/cache/PrisonRegisterStore.ts
```diff
@@ -1,5 +1,5 @@
 import redis from 'redis'
-import type { Prison } from 'prisonTypes'
+import type { PrisonDto } from 'prisonRegisterApiClient'
```
nathanrussell-moj-digital, line 2: I've changed the `PrisonRegisterStore` to cache `PrisonDto`'s rather than our simpler `Prison` type as before. This is because the `PrisonDto` contains the address (`AddressDto`) which we will need later (not in this PR, but later in this story)

server/data/cache/PrisonRegisterStore.ts
```diff
@@ -10,11 +10,11 @@ export default class PrisonRegisterStore extends RedisStore {
     super(redisClient)
   }
 
-  public async setActivePrisons(activePrisons: Array<Prison>, durationDays = 1): Promise<void> {
-    return this.setRedisAsync(ACTIVE_PRISONS, JSON.stringify(activePrisons), 'EX', durationDays * 24 * 60 * 60)
+  public setActivePrisons(activePrisons: Array<PrisonDto>, durationDays = 1) {
```
nathanrussell-moj-digital, line 13: Also changed the signature of `setActivePrisons` to not return the promise - we are not dependant on its return, and we dont wait for its return. It makes the calling code a little cleaner because the linter doesn't issue a warning that we are not doing anything with the return value 👍

server/services/prison/PrisonRegisterService.ts
```diff
@@ -13,12 +13,19 @@ export default class PrisonRegisterService {
   }
 
   async getActivePrisonsFromPrisonRegister(): Promise<Array<Prison>> {
+    let activePrisons: Array<PrisonDto>
```
nathanrussell-moj-digital, line 16: At the moment all I've done is change this public method (`getActivePrisonsFromPrisonRegister`) to use the new type from the cache (`PrisonDto`) and map it into the return type `Prison` (which is our simpler representation containing just an id and name)
At the moment this method is not used as I've not changed the calling code - the calling code still calls the methods that use the static json file.
This is all precursor work to making that switch over

server/services/prison/PrisonRegisterService.ts
```diff
@@ -13,12 +13,19 @@ export default class PrisonRegisterService {
...
     try {
-      const activePrisons = await this.prisonRegisterStore.getActivePrisons()
-      return activePrisons || this.retrieveAndCacheActivePrisons()
+      activePrisons = await this.getActivePrisonDtos()
     } catch (error) {
-      return this.retrieveAndCacheActivePrisons()
+      return error
```
mikehalmamoj, line 20: Possibly something to consider down the line - if we have already cached the prisons and this call is just to do the sporadic update, would we return the stale cached prisons rather than throw an error?

mikehalmamoj (approved):

server/services/prison/PrisonRegisterService.ts
nathanrussell-moj-digital: No, this method is simply about getting the active prisons. The previous code would look to get them from redis and if that returned something a) they would be 'fresh' (ish) as its a 24 hour cache, and b) they would be returned to the caller
If redis returned nothing (null/undefined), then the previous code would call the `retrieveAndCacheActivePrisons` method which would get them from the API and cache them in redis ready for next time.
If either case threw an exception it would attempt to call `retrieveAndCacheActivePrisons` again (!!! - I dont know why I wrote it that way in the first place !!)

The new code still does all that, albeit in a new method `getActivePrisonDtos`. New method does as before (try redis first, else call `retrieveAndCacheActivePrisons`), and just throws if something goes wrong (it has no catch of its own)
The new catch block here returns the thrown error; and that seems to make all the existing tests happy (tests that test exactly this scenario of either redis or the API throwing an error)

Re: the 'sporadic update' we dont really have anything like that per se (and never did have)
I know we have an AC that says prisons will be refreshed after 24 hours - our implementation of that is basically this method. In that with an empty cache, it will get retrieved from prison API and cached with a TTL of 24 hours. Within those 24 hours we will return the cached prisons. Once the TTL has expired redis will evict. At 24 hours and 1 second, we will not re populate the cache. We will have an empty cache until the next time someone requests the prisons, and which point we will get a fresh set of prisons from the API and cache them for a new 24 hour period.
Make sense ?

PR #304, opened 2022-04-26
SLM-234: 404 error handler and nunjucks view

nathanrussell-moj-digital:


integration_tests/integration/page-not-found.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+
+      Page.verifyOnPage(Error404Page)
+    })
+
+    it(`should redirect to the OTC sign in page when not authenticated and navigating to a non existent SLM url`, () => {
+      cy.visit('/barcode/find-recipient-by-inside-leg-measurement', { failOnStatusCode: false })
+
+      Page.verifyOnPage(RequestOneTimeCodePage)
+    })
+  })
```
nathanrussell-moj-digital, line 25: See the change in `nunjucksSetup.ts` 👇 , but there is another use case that we cannot test here
That is where we have an unauthenticated user, and they hit a non-existent url on the SLM domain.
eg: `https://send-legal-mail-to-prisons..../some-non-existent-page`
In that case we would also expect the 404 page, with the SLM page heading
But we cant test that because cypress runs the app up as localhost

server/app.ts
```diff
@@ -105,6 +106,8 @@ export default function createApp(
     setUpCreateBarcode(createBarcodeService, prisonRegisterService, contactService, recipientFormService)
   )
 
+  app.use('/', handleSlm404Errors())
```
nathanrussell-moj-digital, line 109: This is crucial that this is called after the SLM routes are all defined, but before the CR39 authentication and routes

server/errorHandler.ts
```diff
@@ -23,6 +23,9 @@ export default function createErrorHandler(production: boolean) {
 
     res.status(error.status || 500)
 
+    if (res.statusCode === 404) {
+      return res.render('pages/error-404')
+    }
```
nathanrussell-moj-digital, line 28: Errors that are not 404 will still be rendered with the existing error template (which is pretty nasty and ugly); but this should give scope and a way forward for as and when we need to implement the other error content pages

server/utils/nunjucksSetup.ts
```diff
@@ -29,7 +29,9 @@ export default function nunjucksSetup(app: express.Express): void {
       '/cookies-policy',
       '/privacy-policy',
     ]
-    const externalUser = externalUrls.some(externalUrl => req.url.startsWith(externalUrl))
+    const externalUser =
+      externalUrls.some(externalUrl => req.url.startsWith(externalUrl)) ||
+      req.hostname.toLowerCase().startsWith('send-legal-mail-to-prisons')
```
nathanrussell-moj-digital, line 34: This is the line that cannot be tested properly other than by deploying and seeing what happens 🤞

mikehalmamoj (approved):

PR #320, opened 2022-05-11
SLM-245 restore cache before running app for int tests

mikehalmamoj:
This is the fix to a horrible issue we've been suffering where the node app crashed with a segmentation fault when the integration tests started. It appears there was something in the cache that was missing - restoring the cache before running the node app fixed it 🥳


ldlharper (approved):
LGTM

PR #336, opened 2022-05-26
SLM-253 new PrisonService

mikehalmamoj:
* introduce new PrisonService that will become a facade for PrisonRegisterService and SupportedPrisonService
* implement PrisonService in the SupportedPrisonsController
* a future PR will replace all occurences of PrisonRegisterService and config.supportedPrisons


server/app.ts
```diff
@@ -61,7 +62,8 @@ export default function createApp(
   recipientFormService: RecipientFormService,
   zendeskService: ZendeskService,
   cjsmService: CjsmService,
-  supportedPrisonsService: SupportedPrisonsService
+  supportedPrisonsService: SupportedPrisonsService,
```
petergphillips, line 65: this is why I disagree with the typescript template that doesn't allow commas at the end of functions - https://github.com/ministryofjustice/hmpps-template-typescript/blob/main/.eslintrc.json#L93

petergphillips (approved):
LGTM

PR #337, opened 2022-05-27
SLM-253 Make PrisonService the single access point to services/prisons

mikehalmamoj:
This looks big but it's all just replacing PrisonRegisterService and SupportedPrisonsService with the new facade PrisonService.

Once this is merged it'll be easier to replace all the code that uses `config.supportedPrisons` to filter out unsupported prisons.


lbennett-moj (approved):
LGTM

PR #340, opened 2022-05-30
SLM-255 Accessibility audit mitigations 1

mikehalmamoj:


server/filters/renderChooseBarcodeOptionRadiosFilter.ts
```diff
@@ -7,25 +7,25 @@ export default function renderChooseBarcodeOptionRadiosFilter(values: Array<stri
       text: 'Print as part of your own document.',
       hint: {
         html: `        
-          <img src='/assets/images/print-barcodes-image.png' alt='Illustration showing the location of the barcode in the top left corner when inserted into a document' width='94px'/>
```
mikehalmamoj, line 10: I learned from the accessibility audit that as these are non-decorative images and already have instructional text beside them they should be ignored by screen readers - achieved with `alt=""`. Including an alt just doubled up on the instructions for each image which the tester found "frustrating and time consuming".

server/routes/barcode/contacts/updateContactValidator.test.ts
```diff
@@ -76,7 +76,7 @@ describe('updateContactValidator', () => {
     mockValidatePrisonId.mockReturnValue([])
     mockValidatePrisonerDob.mockReturnValue(['some-dob-error'])
 
-    expect(validateContact(form)).toEqual([{ href: '#dob', text: 'some-dob-error' }])
+    expect(validateContact(form)).toEqual([{ href: '#dob-day', text: 'some-dob-error' }])
```
mikehalmamoj, line 79: We chose not to differentiate date errors by day/month/year but this meant that the link in the error header didn't point at an input field. The broken link confused the tester using a screen reader.

mikehalmamoj: There was a separate low priority issue found that suggests we should split out errors by day/month/year which might also mitigate this issue - but I'm still trying to understand how that split could work without getting very complicated.

server/views/pages/barcode/edit-contact-details.njk
```diff
@@ -47,7 +47,6 @@
               name: "prisonerName",
               value: form.prisonerName,
               attributes: {
-                autofocus: '',
```
mikehalmamoj, line 50: Autofocus is a bad thing for accessibility - it drops users right in the middle of the page meaning they lose all context of where they are.

joelstobart-moj (approved):
LGTM

Interesting feedback!

PR #355, opened 2022-06-25
SLM:261: Improve error handling in scan barcode routes

tpmcgowan:
Add async middleware to ensure unexpected API responses from SLM backend (e.g. 401/403) are ultimately passed to error handler rather than crashing the application.


psoleckimoj (approved):
LGTM!

PR #363, opened 2022-08-23
Update ingress.yaml as prerequisite to switching ingress to new controller.

dhirajshettymoj:
Update ingress.yaml file as mentioned here - https://user-guide.cloud-platform.service.justice.gov.uk/documentation/networking/apiversion-changes-ingress-k8s-1-22.html as part of prerequisite for switching ingress to the new nginx-ingress-controller v1.2.


tpmcgowan (approved):
LGTM

PR #372, opened 2022-10-03
Feature/SLM-275 changes to start page

dhirajshettymoj:
SLM-275 changes- 
Update to start page to replace list of prisons with generic text.
Minor dependency updates.


server/views/pages/start/legal-sender-start-page.test.ts
```diff
@@ -21,12 +20,12 @@ describe('Legal Sender Start Page View', () => {
...
-  it('should show supported prison names', () => {
-    viewContext = { prisonNames: ['some-prison', 'another-prison'] }
-
-    const $ = cheerio.load(compiledTemplate.render(viewContext))
-
-    expect($('[data-qa=prisonNames]').html()).toContain('some-prison')
-    expect($('[data-qa=prisonNames]').html()).toContain('another-prison')
+  it('should show updated any prison supported text instead of prison names', () => {
+    const $ = cheerio.load(compiledTemplate.render({}))
+    const startNowHrefSelector = 'a[href="/oneTimeCode/request-code"]'
```
tpmcgowan, line 25: If you add an attribute to the relevant `h2` in the `.njk` file (e.g. `<h2 class="govuk-heading-m" data-qa="prisons">Prisons you can send mail to</h2>`) then you could just target it directly like this:

```javascript
expect($('[data-qa="prisons"]').text()).toEqual('Prisons you can send mail to')
expect($('[data-qa="prisons"] + p').text()).toContain(
  'You can use this service to send legal or confidential mail to any prison'
)
```

([cheerio](https://cheerio.js.org) uses jQuery style selectors so you don't necessarily need to add custom attributes if there's already a class or id you can target.)

server/views/pages/start/legal-sender-start-page.njk
```diff
@@ -51,12 +51,7 @@
             }) }}
 
             <h2 class="govuk-heading-m">Prisons you can send mail to</h2>
-            <p>You can use this service to send legal or confidential mail to the following prisons:</p>
-            <ul class="govuk-list govuk-list--bullet" data-qa="prisonNames">
-            {% for prisonName in prisonNames %}
```
tpmcgowan, line 56: `prisonName` data (and the API call behind it) is still getting passed to this template even though you're not displaying it. See here:https://github.com/ministryofjustice/send-legal-mail-to-prisons/blob/fe0388dec1495eb476609de6e741eaaac945eb73/server/routes/start/StartPageController.ts#L12

You could get rid of most of the lines in that file as you just need to render the template; no need to do API call. (Might then need to tidy up a bit as `prisonService` would no longer need to be passed)

server/views/pages/start/legal-sender-start-page.test.ts
dhirajshettymoj: Updated both page and test based on your recommendations. Thanks.

server/views/pages/start/legal-sender-start-page.njk
dhirajshettymoj: Good spot, thanks. Updated StartPageController.ts to remove the call to PrisonService as advised.

tpmcgowan (approved):
LGTM!

PR #376, opened 2022-11-17
Update typescript dependency version.

dhirajshettymoj:
Update to typescript dependency version as a fix for security failures.


tpmcgowan (approved):
LGTM!

PR #391, opened 2023-02-10
SLM-270: Fix smoke tests in CircleCI

tpmcgowan:
* Fix installation of `jq` dependency that was failing
* Remove unnecessary `baseUrl` config that was incorrectly overriding CI config
* Fix error in `run-smoke-tests.sh` that stopped build directory being deleted locally


dhirajshettymoj (approved):
LGTM

PR #408, opened 2023-07-12
Dependency updates.

dhirajshettymoj:
Dependency updates as a fix for security failures.


server/utils/azureAppInsights.test.ts
```diff
@@ -26,7 +26,7 @@ const createContext = (username: string, activeCaseLoadId: string) =>
         },
       },
     },
-  } as ContextObject)
+  }) as ContextObject
```
dhirajshettymoj, line 29: Change as suggested by eslint. @psoleckimoj @tpmcgowan could you please check if this is correct?

server/data/hmppsAuthClient.ts
```diff
@@ -48,7 +48,9 @@ export interface UserRole {
 }
 
 export default class HmppsAuthClient {
-  constructor(private readonly tokenStore: TokenStore) {}
+  constructor(private readonly tokenStore: TokenStore) {
+    // do nothing
+  }
```
tpmcgowan, line 53: Why have all these appeared? Is it some new default settings...not sure it's an improvement!

dhirajshettymoj: I updated the eslint version. I think the new eslint version flags for empty methods

psoleckimoj: @dhirajshettymoj @tpmcgowan just thinking it might be worth checking the eslint rules on the template project to see if they have new rules and/or exclusions as I remember seeing this happening.

psoleckimoj (dismissed):
LGTM! Made a comment re: asking about whether to disable this rule or not, think it's more a community thing.

server/data/hmppsAuthClient.ts
tpmcgowan: Just had a quick look on another project (so haven't tried here) but  I think you can stop it wanting to complain about empty constructors by adding the following:
```
no-empty-function: ["error", { "allow": ["constructors"] }]
```
to the `rules` section here: https://github.com/ministryofjustice/send-legal-mail-to-prisons/blob/4019d6a33d10ae4cd4e06da2a34415dff273e203/.eslintrc.json#L36

dhirajshettymoj: thanks @tpmcgowan . @psoleckimoj also suggested doing the same.

server/utils/azureAppInsights.test.ts
psoleckimoj: Run it through `npm run typecheck` and it'll make sure TypeScript is happy 👍

server/data/hmppsAuthClient.ts
dhirajshettymoj: I have updated the PR now based on your suggestions.

server/utils/azureAppInsights.test.ts
dhirajshettymoj: thanks.. no issues spotted

psoleckimoj (approved):
LGTM!

tpmcgowan (approved):
LGTM

PR #411, opened 2023-10-02
Update local ingresses on dev

dhirajshettymoj:
Update local ingresses on dev to remove references to staff facing application and keep only public facing application URL.


helm_deploy/send-legal-mail-to-prisons/values.yaml
```diff
@@ -10,7 +10,14 @@ generic-service:
...
 
   ingress:
-    enabled: false
+    enabled: true
+    host: app-hostname.local    # override per environment
+    tlsSecretName: send-legal-mail-cert
+    annotations:
+      nginx.ingress.kubernetes.io/configuration-snippet: |
+        location = / {
+          return 301 $scheme://$host/barcode/find-recipient;
```
dhirajshettymoj, line 19: Copied from existing ingress. Redirects to /barcode/find-recipient if the user lands up on /

psoleckimoj (approved):
LGTM!

helm_deploy/send-legal-mail-to-prisons/values.yaml
tpmcgowan: Could put this as a redirect in the Node app itself? Perhaps might need to wait until the unneeded 'other app' code is stripped out (which I guess is why it was configured here in the first place...)

tpmcgowan (approved):
LGTM

PR #412, opened 2023-10-12
Remove expectHelpdeskLink from pages on new staff UI 

dhirajshettymoj:
Set expectHelpdeskLink to false on pages used by new staff UI as smoke tests fail since helpdesk links are no longer available on staff UI.


tpmcgowan (approved):
LGTM

PR #426, opened 2024-02-15
Slm 340 update gov.uk header on public service

simonnebMOJ:
I have had a go at updating the Header for the new logo.


server/views/partials/header.njk
```diff
@@ -1,45 +1,39 @@
...
-            </a>
-          </li>
-
-          <li class="hmpps-header__navigation__item">
-            <a data-qa="signOut" class="hmpps-header__link" href="/sign-out">Sign out</a>
-          </li>
-        {% endif %}
-
+    <div class="govuk-header__content">
+      <a href="/" class="govuk-header__link govuk-header__service-name">
```
dhirajshettymoj, line 20: can we add data-qa="header-text" to this tag as a fix for a failing integration test.

dhirajshettymoj, line 20: can we add data-qa="header-text" to this tag as a fix for a failing integration test.

simonnebMOJ: Thanks - updated. Integration tests now passing.

dhirajshettymoj (approved):
LGTM

PR #431, opened 2024-02-26
remove smoke tests temporarily since we haven't got a fix for the cer…

dhirajshettymoj:
remove smoke tests temporarily since we haven't got a fix for the certificate issue yet. Brief detail about the issue - 

Since the last few days we have been getting this error while trying to run the smoke tests. This is primarily down to the apt-get update failing to update the certs. Tried multiple options however but seems to fail every time. Also updated the dockerfile from bullseye to bookworm but still the same issue. So disabling the smoke tests till we get around this issue. Hopefully Google will update the certs post which it starts running again as earlier - 

Example - 
https://app.circleci.com/pipelines/github/ministryofjustice/send-legal-mail-to-prisons/1992/workflows/df2d6d09-6383-4ff9-8259-a2cb3deda3b8/jobs/9233

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ feature_branch: &feature_branch
 version: 2.1
 
 orbs:
-  hmpps: ministryofjustice/hmpps@7
+  hmpps: ministryofjustice/hmpps@8
   slack: circleci/slack@4.12.1
 
 parameters:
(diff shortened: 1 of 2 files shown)


tpmcgowan:
Have you tried updating the executor for the smoke tests. It's using an old version: https://github.com/ministryofjustice/send-legal-mail-to-prisons/blob/2b2f1836322523a840df9e190313b2957a214122/.circleci/config.yml#L32

Newer versions here: https://hub.docker.com/r/cypress/included. E.g. could match it to what's in `package.json`. 

Vaguely recall this was what #390 was about...

PR #481, opened 2024-09-18
Upgrade to redis 4/connect-redis 7

psoleckimoj:


server/data/cache/SmokeTestStore.ts
```diff
@@ -1,35 +1,64 @@
...
+    if (!req.body?.msjSecret || !config.smoketest.msjSecret) {
+      return ''
+    }
+
     const secret = req.body.msjSecret
-    if (secret === config.smoketest.msjSecret) {
-      const oneTimeSecret = crypto.randomBytes(20).toString('hex')
-      await this.setSmokeTestSecret(oneTimeSecret)
-      return oneTimeSecret
+
```
psoleckimoj, line 53: Just flipped the order to make it neater

server/routes/barcode/pdf/PdfController.test.ts
```diff
@@ -304,7 +304,6 @@ describe('PdfController', () => {
 
       try {
         await pdfController.downloadPdf(req as unknown as Request, res as unknown as Response)
-        fail('Was expecting pdfController.downloadPdf to have thrown an error but it did not')
```
psoleckimoj, line 307: `fail` was an internal, undocumented jest function and should never have been used. It is not present in newer jest versions. All this seems to be used for is presenting a message.

tpmcgowan (approved):
LGTM!

PR #608, opened 2025-08-28
Feature/move security scans to gha

psoleckimoj:


dhirajshettymoj (dismissed):
LGTM. Just 1 question, scans for both apps will run at the same time. do they need to be configured to run at different times?

dhirajshettymoj (approved):
LGTM

